PR #19, opened 2022-09-21
Clean up types and add more CSS

keithamus:
This PR installs typescript so we can use the type checking to fix a few small bugs around improper types. As well as that I've updated the CSS to fall in-line with the UA CSS in Chrome

## Steps to test/reproduce
Running `npm run lint:js` will show you typescript running and passing.

## Show me
The CSS now centres the popup in the middle of the page.

a screenshot of the index.html popups, centred in the page


tsconfig.json
```diff
@@ -0,0 +1,13 @@
+{
```
sanajaved7, line 1: FYI @keithamus we updated the [polyfill repo template](https://github.com/oddbird/polyfill-template) but haven't yet pulled in all the new changes into this repo. There are quite a number of new things and changes so I'll let you decide if you would like to pull those in or if it would be too much. Feel free to use the template repo as a reference and pull things as you need.

sanajaved7 (approved):

tsconfig.json
keithamus: I’ll update this tomorrow to include the changes from the template!

PR #36, opened 2022-11-14
rename popup to popover

keithamus:
## Description

The [latest spec changes](https://whatpr.org/html/8221/popover.html#the-popover-attribute) have renamed the `popup` attribute to `popover`. Consequently this PR will need to rename all instances of `popup` to `popover`. This PR changes all references within the code, which includes ts, css and tests.

I've avoided changing references in the `package.json` as that reflects the repository name too. Perhaps someone from Oddbird can think about the correct course here? My vote would be to rename this repo `popover-polyfill`, as GitHub will redirect the old name (so long as a repository named `oddbird/popup-polyfill` does not exist to replace it).


src/popover.ts
```diff
@@ -105,54 +105,54 @@ export function apply() {
...
-    for (const popup of doc.querySelectorAll(
-      '[popup="" i].\\:open, [popup=auto i].\\:open, [popup=hint i].\\:open',
+    // Dismiss open Popovers
+    for (const popover of doc.querySelectorAll(
+      '[popover="" i].\\:open, [popover=auto i].\\:open, [popover=hint i].\\:open',
     )) {
-      if (popup instanceof HTMLElement && popup !== effectedPopup)
-        popup.hidePopUp();
+      if (Popover instanceof HTMLElement && Popover !== effectedPopover)
+        Popover.hidePopover();
```
jgerigmeyer, line 155:
```suggestion
      if (popover instanceof HTMLElement && popover !== effectedPopover)
        popover.hidePopover();
```

keithamus: :+1:

jgerigmeyer (approved):
@keithamus Looks good! I agree with you re: renaming the repo. I'll open a new PR with that change, as well as some cleanup to the docs and build process.

PR #40, opened 2022-11-17
Fix event handling of buttons

keithamus:
## Description

`<button>` tags can have HTML inside of them. Currently the code for button handling checks to see if `event.target` is an appropriate button element, but this isn't accurate because the `event.target` will be the element that initiated the event, which could be the inner HTML of a button element.

In order to validate this, I've added some `<span>` elements inside the `<button>` tags, which fails the test suite as the buttons no longer function. In a separate commit I have corrected the event handler code to grab the closest applicable button element in the tree which has one of the `popoverXtarget` attributes.

## Steps to test/reproduce
If you checkout commit [de6566c](https://github.com/oddbird/popover-polyfill/pull/40/commits/de6566c28b9bcc719b76bc79e8cd670fa8dfaca6) you can validate that "Click to show Popover 5" and "Click to toggle Popover 9" won't work. Checking out HEAD of this branch will work.

### DIFF
--- a/index.html
+++ b/index.html
@@ -28,9 +28,13 @@
       <button popovertoggletarget="popover1">Click to toggle Popover 1</button>
       <button popovertoggletarget="popover2">Click to toggle Popover 2</button>
       <button popovershowtarget="popover3">Click to show Popover 3</button>
-      <button popovershowtarget="popover5">Click to show Popover 5</button>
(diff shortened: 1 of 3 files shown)


jgerigmeyer (approved):
🎉

PR #49, opened 2022-12-20
fix(popover): support shadow tree elements lookup

### LINKED ISSUE #44: Using popover inside shadowRoot
Hi!

We are trying to use popover polyfill within custom elements.

Since the popover attribute is inside the shadow root, it cannot be found.

This is a screenshot from `popover.js`, `doc.querySelectorAll('[popover="" i].\\:open, [popover=auto i].\\:open')`.

Screenshot 2022-12-06 at 16 11 30

yinonov:
querySelector looking for all `[popover]` attributes in a document will now pierce within
shadow trees recursively.

this introduces a use of a package [query-selector-shadow-dom](https://www.npmjs.com/package/query-selector-shadow-dom).

What bugs me with this solution is that it the package `querySelectorAllDeep` method doesn't operate from a specific element, and rather the document element, thus removes the explicit query from `ownerDocument` of the polyfill (might be good to raise a request for the team [query-selector-shadow-dom/issues](https://github.com/webdriverio/query-selector-shadow-dom/issues)).

this probably has little effect in browsers, as, commonly, `ownerDocument` will always point at the document element.

would love to hear maintainers' thoughts on this...

Closes #44

issues to resolve -
- [x] declarative shadow root doesn't trigger attachShadow method, therefore, its template content doesn't get picked up. maybe override `connectedCallback` instead. can the static method `observedAttributes` assist to trigger popover attributes change?
- [x] css doesn't pierce in to shadowRoot, therefore not hiding popovers. should a caveat be added to state "does not work apply to shadow Dom" in
https://github.com/yinonov/popover-polyfill/blob/8bac7a2d5b74428c75ac08696ff197747722c72d/README.md#L29-L33
?
- [ ] [document.addEventListener('click',](https://github.com/yinonov/popover-polyfill/blob/37c852d856b6214fd64c8b3424dd95f23816e6af/src/popover.ts#L102) doesn't trigger actions inside various shadow roots
- [ ] identical IDs in different shadow roots may conflict looking up `effectedPopover` and return a wrong element. should relations between popover button and its control NOT bleed out of same `ownerDocument`?


jgerigmeyer:
@yinonov Thanks for the contribution! I'd love to hear what @keithamus thinks about this before merging.

keithamus:
I'm a little bit concerned looking through the query selector library, as it looks like it might cause slowdowns on pages with lots of shadow roots. It also looks like it won't work on closed shadowRoots.

Given we know we only want to interact with elements that have a `popover` attribute, I wonder if it would be more beneficial to track just those elements with a MutationObserver.

To use a MutationObserver we could patch `attachShadow` to get references to shadowRoots and add them to the observer, tracking for child elements that get the `popover` attribute. The problem with that, however, would be declarative shadowRoots.

yinonov:
> I'm a little bit concerned looking through the query selector library, as it looks like it might cause slowdowns on pages with lots of shadow roots. It also looks like it won't work on closed shadowRoots.
> 
> Given we know we only want to interact with elements that have a `popover` attribute, I wonder if it would be more beneficial to track just those elements with a MutationObserver.
> 
> To use a MutationObserver we could patch `attachShadow` to get references to shadowRoots and add them to the observer, tracking for child elements that get the `popover` attribute. The problem with that, however, would be declarative shadowRoots.

given the [lack of broad support](https://caniuse.com/declarative-shadow-dom) for declarative shadowRoots, can the `attachShadow` monkey patching and observation considered as first aid?

keithamus:
It might be worth looking into how declarative shadow roots are assigned to see if it would significantly alter the architecture. Having said that I have a hunch it won't be easy and may require extensive modifications to multiple APIs, and so we might be better off ignoring them for now. 

Closed shadow roots are only accessible via `attachInternals().shadowRoot` but that's not always going to be called, and simple polyfills for DSD dont polyfill attachInternals. Open shadowRoots will assign `.shadowRoot` but I have a hunch it will do so using `Object.defineProperty` which is not observable without significant workarounds such as patching defineProperty itself, or using Proxies which have a performance penalty.

I think it's quite reasonable to add "does not work with declarative shadow Dom" to the list of caveats for now. That means the only mechanism for creating shadow roots is with `attachShadow` which makes for a more straightforward patch.

yinonov:
image

I'm up for the _"does not work with declarative shadow Dom"_ caveat as, I believe, is not so very common yet.
boiled something up with patching `attachShadow`. will push soon

yinonov:
**note to self** - verify the following pitfall
> If you call observe() on a node that's already being observed by the same MutationObserver, all existing observers are automatically removed from all targets being observed before the new observer is activated.

will any shadowRoots within shadowRoots remove parent observation?

yinonov:
as this seem to be bigger than the initial expectation, I'd ask for a short review before moving forward.

The plan is to move new code to a class that handles all observation of document or shadowRoot.

removed the `query-selector-shadow-dom`.

src/popover.ts
```diff
@@ -179,7 +186,10 @@ export function apply() {
     const target = event.target;
     if (!(target instanceof Element)) return;
     const doc = target.ownerDocument;
-    let effectedPopover = closestComposed(event);
+    let effectedPopover = (event
+      .composedPath()
```
keithamus, line 88: IIRC `composedPath` does not work with closed shadow roots. It may be worth adding tests for both open and closed shadow roots for this.

yinonov: refactored `composedPath` to a custom `closest` function. didn't override the prototype `closest` to keep consumers' environment intact

tests/dismiss.spec.ts
```diff
@@ -15,13 +15,16 @@ test('click dismisses all auto popovers', async ({ page }) => {
...
   const popover10 = (await page.locator('#popover10')).nth(0);
+  const popover11 = (await page.locator('#popover11')).nth(0);
   await expect(popover10).toBeHidden();
 
   await page.click('h1');
   await expect(popover7).toBeHidden();
   await expect(popover8).toBeHidden();
   await expect(popover9).toBeHidden();
   await expect(popover10).toBeHidden();
+  await expect(popover10).toBeHidden();
```
yinonov, line 26:
```suggestion
```

keithamus (commented):
I’ve briefly read through the PR just now and it looks to be shaping up really nicely! I’ll take a closer look next week to offer some more substantial feedback but I’d say so far so good!

yinonov:
couple new issues to resolve -

> * [ ]  declarative shadow root doesn't trigger attachShadow method, therefore, its template content doesn't get picked up. maybe override `connectedCallback` instead. can the static method `observedAttributes` assist to trigger popover attributes change?
> * [ ]  css doesn't pierce in to shadowRoot, therefore not hiding popovers. should a caveat be added to state "does not work apply to shadow Dom" in
>   https://github.com/yinonov/popover-polyfill/blob/8bac7a2d5b74428c75ac08696ff197747722c72d/README.md#L29-L33
>   ?

yinonov:
triggering elements same tree scope limitation - openui/open-ui#651

yinonov:
popover triggering elements should operate on their own tree scope but given that the polyfill only triggers [document.addEventListener('click'](https://github.com/yinonov/popover-polyfill/blob/37c852d856b6214fd64c8b3424dd95f23816e6af/src/popover.ts#L102) events fired only refer target to `host`.
when clicking on triggering elements within a shadow root will only fire with target of host, causing the likes of [button.hasAttribute('popovershowtarget')](https://github.com/yinonov/popover-polyfill/blob/37c852d856b6214fd64c8b3424dd95f23816e6af/src/popover.ts#L116) to always return false. the attributes (`popovertoggletarget` | `popoverhidetarget` | `popovershowtarget'`) don't match target.

should each tree have its own listener? is there a way to intercept any click on popover triggering elements no matter the tree scope?

keithamus:
Each tree will need its own event listener, or you could perhaps check the first element of `composedPath` - though I think that might not work for closed shadow roots.

yinonov:
> Each tree will need its own event listener, or you could perhaps check the first element of `composedPath` - though I think that might not work for closed shadow roots.

indeed, `composedPath` wont do for closed shadow roots. will try each tree. would evenListeners get garbage collected when removed?

keithamus:
> Would evenListeners get garbage collected when removed?

Yes, if you imagine each `EventTarget` instance (of which ShadowRoots are one) conceptually has a `#listeners = new Set<Function>()` private property, you can imagine how when an EventTarget is garbage collected, then all listeners are also collected (provided no hard references are held elsewhere).

src/popover.ts
```diff
@@ -9,6 +11,37 @@ export function isSupported() {
...
+
+function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
+  Element.prototype[ORIGINAL_ATTACH_SHADOW_SYMBOL] =
+    Element.prototype.attachShadow;
+  Element.prototype.attachShadow = function (init) {
+    const shadowRoot = this[ORIGINAL_ATTACH_SHADOW_SYMBOL](init);
+    callback(shadowRoot);
+    return shadowRoot;
+  };
+}
```
keithamus, line 24: You could use a variable to capture this, and avoid having to add extra fields to Element.prototype which in turn needs changes to TS:

```suggestion
export const originalAttachShadow = Element.prototype.attachShadow

function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
  Element.prototype.attachShadow = function (init) {
    const shadowRoot = originalAttachShadow.call(this, init);
    callback(shadowRoot);
    return shadowRoot;
  };
}
```

yinonov: strangely this simple solution failed when I initially set this up. well, now it charmlessly works

yinonov: scoped the const within function to allow multi patching

README.md
```diff
@@ -36,6 +36,9 @@ with a `<link rel=stylesheet>` tag:
 <link rel="stylesheet" src="/path/to/popover.css" />
 ```
 
+Note that default styles will not be applied to shadow roots.
+Each shadow tree will need to include the styles explicitly.
```
yinonov, line 40:
```suggestion
Each root will need to include the styles explicitly.
```

yinonov:
Should this be progressed? it doesn't seem like the repo is frequently maintained. are there too many technical limitations to this end goal? is it the lack of resolved cracks in specification?

keithamus:
@yinonov please do progress with it! The repo is maintained and the polyfill is actively used. 

Do you feel like this PR is now in a good place to review? I've been holding off on a full review, as I wanted you to feel comfortable with the code being ready for review.

src/observar.ts
```diff
@@ -0,0 +1,66 @@
+export const popovers = new Set<HTMLElement>();
+
+const handleChildListMutation = (mutation: MutationRecord) => {
+  if (mutation.addedNodes.length > 0) {
+    mutation.addedNodes.forEach((node) => {
+      if (node instanceof HTMLElement) {
+        if (node.hasAttribute('popover')) {
```
keithamus, line 7: I think you might also need to add `quetySelectorAll('[popover]')` here too. Might be worth adding a test for nodes that get added that have a popover element as a child.

src/observar.ts
```diff
@@ -0,0 +1,66 @@
...
+          popovers.add(node);
+        }
+      }
+    });
+  }
+
+  mutation.removedNodes.forEach((node) => {
+    if (node instanceof HTMLElement && node.hasAttribute('popover')) {
```
keithamus, line 15: I think you might also need to add `quetySelectorAll('[popover]')` here too. Might be worth adding a test for nodes that get removed that have a popover element as a child.

keithamus (approved):
This looks like a very nice implementation. I just have one small concern about nested popovers that happen during a mutation.

src/popover.ts
```diff
@@ -1,3 +1,5 @@
+import { observePopoversMutations, popovers } from './observar.js';
```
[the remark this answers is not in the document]
yinonov: no, it's intentionally a verb not adjective

src/observar.ts
yinonov: good catch!

yinonov: done

src/popover.ts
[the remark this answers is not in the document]
yinonov: doesn't make sense?

yinonov:
I keep getting `Error: Timed out waiting 10000ms from config.webServer.` when running `npm test`. any idea how to resolve? I followed the [contribution guide](CONTRIBUTING.md) precisely

jgerigmeyer:
> I keep getting `Error: Timed out waiting 10000ms from config.webServer.` when running `npm test`. any idea how to resolve? I followed the [contribution guide](/oddbird/popover-polyfill/blob/main/CONTRIBUTING.md) precisely

@yinonov I've hit that recently too, and it seems to be inconsistent for me -- it passes once, then times out. Something in the connection between playwright and esbuild, but I haven't had time to dig into it yet.

src/popover.ts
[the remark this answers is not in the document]
keithamus: I am not sure `observar` is the English noun, I think `observer` is both the adjective and noun.

keithamus (approved):
This looks really good to me now 👍

yinonov:
> > I keep getting `Error: Timed out waiting 10000ms from config.webServer.` when running `npm test`. any idea how to resolve? I followed the [contribution guide](/oddbird/popover-polyfill/blob/main/CONTRIBUTING.md) precisely
> 
> @yinonov I've hit that recently too, and it seems to be inconsistent for me -- it passes once, then times out. Something in the connection between playwright and esbuild, but I haven't had time to dig into it yet.



> This looks really good to me now 👍

so is it possible to at least run tests in CI pre-merge?

jgerigmeyer:
> so is it possible to at least run tests in CI pre-merge?

We could merge this into a branch and allow CI to run there before merging into `main`...

I can't reproduce the timeout locally, only occasionally on CI. Locally, can you run `npm run serve` in one terminal, then comment out the `webServer` settings in `playwright.config.ts` and run `npm run test:ci`?

yinonov:
> webServer

thanks, that did it for now. resolving failures

jgerigmeyer:
I think there's a way we can modify the `test` workflow to run on public fork PRs, but I haven't looked into it yet. @keithamus?

keithamus:
I think it might be this line stopping them running:  https://github.com/oddbird/popover-polyfill/blob/dc20d5aab11dd1e9d479aedb9dedb3b3026a3fd7/.github/workflows/test.yml#L6

yinonov:
facing an issue with playwright [locator not supporting close mode shadowRoot](https://playwright.dev/docs/locators#locate-in-shadow-dom)

image

tests pass when setting [host.attachShadow({ mode: 'closed' });](https://github.com/yinonov/popover-polyfill/blob/4d03147f3819b2c906370f7686e5fd3b76a81c63/index.html#L30) to `'open'`

we still want to ensure close shadowRoots are tested, don't we?

yinonov:
> facing an issue with playwright [locator not supporting close mode shadowRoot](https://playwright.dev/docs/locators#locate-in-shadow-dom)
> 
> image
> 
> tests won't pass unless [host.attachShadow({ mode: 'closed' });](https://github.com/yinonov/popover-polyfill/blob/4d03147f3819b2c906370f7686e5fd3b76a81c63/index.html#L30) is set to `'open'`
> 
> we still want to ensure close shadowRoots are tested, don't we?

I don't see a way around this blocker unless heavily extending the testing environment. don't feel right.
`'close'` shadowRoot content is unreachable, as it is intended to be in browsers

src/observar.ts
```diff
@@ -0,0 +1,77 @@
...
+        break;
+    }
+  });
+};
+
+const observer = new MutationObserver(handleMutation);
+
+export const observePopoversMutations = (root: Document | ShadowRoot) => {
+  // Document don't initially trigger childList mutations as opposed
+  // to shadow roots. so we need to manually add the popovers to the set
```
jgerigmeyer, line 63:
```suggestion
  // Documents don't initially trigger childList mutations as opposed
  // to shadow roots, so we need to manually add the popovers to the set
```

jgerigmeyer (changes requested):
Functionality looks good! A few minor changes, then I think this is ready to merge.

index.html
```diff
@@ -22,6 +23,23 @@ <h1>Popover Attribute Polyfill</h1>
...
+        shadowRoot.innerHTML = `<button popovertoggletarget="shadowedPopover">
+                  Click to toggle shadowed Popover (auto)
+                </button>
+                <button popovertoggletarget="shadowedNestedPopover">
+                  Click to toggle shadowed nested Popover (auto)
+                </button>
+                <div id="shadowedPopover" popover="auto">Shadowed Popover (auto)</div>
+                <div>
+                  <div id="shadowedNestedPopover" popover="auto">Shadowed Nested Popover (auto)</div>
+                <div>`;
```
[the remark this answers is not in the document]
yinonov: good catch

src/popover.ts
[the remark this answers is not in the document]
yinonov: Was just a typo 🤦🏼‍♂️

jgerigmeyer (approved):
🚀

yinonov:
@jgerigmeyer @keithamus thank you for all the support!

jgerigmeyer:
@yinonov Thanks for your contribution! I'll make a new release today or Monday.

jgerigmeyer:
Published [v0.0.5](https://github.com/oddbird/popover-polyfill/releases/tag/v0.0.5) :shipit:

PR #52, opened 2023-01-05
docs: even also include CSS within the CDN section

mfranzke:
## Description
As it's mentioned in the [CDN section](https://github.com/oddbird/popover-polyfill/blob/main/README.md?plain=1#L58) this part would be for prototyping or testing, so we could mention the CSS in here directly as well, even though that it's separated in the previous sections that are meant for production use and have some further description regarding CSS usage.


keithamus (approved):

keithamus:
Thanks @mfranzke! Nice catch!

PR #55, opened 2023-01-12
use standard text and background colors

nuxodin:
The system colors are supported in every recent browser version: `canvastext` `canvas`

I don't know if we still need the fallbacks, or which browsers the polyfill supports. Otherwise we could do without the media-query.

## Steps to test/reproduce

## Show me

### DIFF
--- a/src/popover.css
+++ b/src/popover.css
@@ -11,6 +11,8 @@
   border: 1px solid;
   background: white;
   color: black;
+  background: canvas;
+  color: canvastext;
   overflow: auto;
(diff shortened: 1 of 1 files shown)


keithamus (approved):
Cool I never knew about these properties! According to [MDN they've been supported since forever](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color) so I'd be happy to drop the media query too!

jgerigmeyer (approved):

PR #60, opened 2023-01-16
Fixup css

keithamus:
## Description

This changes up some of the CSS to align closer to [Chrome's user-agent CSS](https://github.com/chromium/chromium/blob/main/third_party/blink/renderer/core/css/popover.css). What's changed:

 - Simplify selectors to just `[popover]` as Chrome does.
 - Remove `top`, `left` in favour of `inset`.
 - Drop the `inset-*-*` properties.
 - Shrink padding from `1em` to `0.25em`.
 - Drop the 1px default border. Chrome's border-width is `initial`.
 - Move the z-index into the `[popover]` selector so it's always applied
 - Rather than `[popover]` having `display: none`, and `[popver]:open` being `display: block`, we now have `[popover]:not(:open)` as `display: none`. This way one can style a popover element to be something like `display: grid` or `display: flex` and popover will not override that.
 - Add `[popover][anchor]` styles, to avoid conflicting with anchored positioning.

## Steps to test/reproduce

## Show me

### DIFF
--- a/tests/basic.spec.ts
+++ b/tests/basic.spec.ts
@@ -44,7 +44,6 @@ test('popover as "manual"', async ({ page }) => {
 
 test('popover as "invalid"', async ({ page }) => {
   const popover = (await page.locator('#popover6')).nth(0);
-  await expect(popover).toBeVisible();
   await expect(
     async () => await popover.evaluate((node) => node.showPopover()),
   ).rejects.toThrow(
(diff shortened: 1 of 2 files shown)


tests/basic.spec.ts
```diff
@@ -44,7 +44,6 @@ test('popover as "manual"', async ({ page }) => {
 
 test('popover as "invalid"', async ({ page }) => {
   const popover = (await page.locator('#popover6')).nth(0);
-  await expect(popover).toBeVisible();
```
keithamus, line 47: This expectation has been dropped as it does not hold true in a default Chrome environment.

jgerigmeyer (approved):

PR #68, opened 2023-01-26
add beforetoggle events

keithamus:
## Description

This PR adds the `beforetoggle` event which is defined by the spec: https://whatpr.org/html/8221/popover.html#show-popover. It also exposes the global `BeforeToggleEvent` which is a new `Event` subclass that is exposed on `window`.

The `beforetoggle` event has two properties: `currentState` and `newState`. They are set to `'open'` or `'closed'` based on the current and desired state of the popover. The `beforetoggle` event when _showing_ a popover is cancelable. The event when hiding a popover is _not_ cancelable. This event does not bubble.

## Steps to test/reproduce

I've added tests for each of these conditions. Check out `events.spec.ts`

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@oddbird/popover-polyfill",
-  "version": "0.0.7",
+  "version": "0.0.8",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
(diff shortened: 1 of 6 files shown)


keithamus:
I've also added the `0.0.8` version commit, so once merged this can be published as such.

jgerigmeyer (approved):
Looks great! 🚀

PR #70, opened 2023-01-27
feat(popover): add support for popover target element

### LINKED ISSUE #66: Support target element reference
any hints on strategy for adding support to JS target element reference?
https://github.com/openui/open-ui/issues/651#issuecomment-1397519303

yinonov:
Closes #66

Note that this PR:
- Refactored popover triggers/invokers handling
- Had to deviate and extend from the [spec exact flow](https://whatpr.org/html/8221/popover.html#popover-target-element) due to the lack of properties reference (e.g. 
> If node has a [popovertoggletarget](https://whatpr.org/html/8221/popover.html#attr-popover-toggle-target) attribute, then set idref to the value of node's [popovertoggletarget](https://whatpr.org/html/8221/popover.html#attr-popover-toggle-target) attribute

doesn't mentioned property check for `popoverToggleTargetElement`
- introduced another `triggers.ts` file to extract payload of code from the main `popover.ts`
- `popover.ts` is still cumbersome and already hard to read
- [x] should organize code more neatly
- [x] should write more tests
- [ ] remove comments


yinonov:
@keithamus @jgerigmeyer 
please review, as far as functionality. code runs as expected.
I'm yet sure about the style of code, it seem to bloat rapidly. WDYT?

src/popover.ts
```diff
@@ -227,4 +206,40 @@ export function apply() {
...
+    set: function (el: unknown) {
+      if (!(el instanceof Element)) {
+        throw new TypeError('popoverToggleTargetElement must be an element');
+      }
+      this._popoverToggleTargetElement = new WeakRef(el);
+    },
+    get: function () {
+      return this._popoverToggleTargetElement?.deref() || null;
+    },
+  });
```
keithamus, line 220: A few of points on this pattern:

Using `_` allows for privacy by convention but not enforced by the language. We should avoid this and instead enforce it with the language by using a WeakMap.

We're missing the concept that these must _reflect_ the attribute. We need to query the DOM if these aren't set with a weakmap.

While `WeakRef` support is pretty good, it's not possible to polyfill it (any implementation would hold a strong reference to `el`). This means we're essentially tying support to browsers that support `WeakRef`. One way around this is to only return connected elements, and explicitly drop the reference if the element isn't connected.

So a refactor might look like:

```suggestion
  const togglesByEl = new WeakMap<Element, Element>()
  Object.defineProperty(HTMLElement.prototype, 'popoverToggleTargetElement', {
    set(el: unknown) {
      if (!(el instanceof Element)) {
        throw new TypeError('popoverToggleTargetElement must be an element');
      }
      togglesByEl.set(this, el)
    },
    get() {
      const el = togglesByEl.get(this)
      if (!el?.isConnected) {
        togglesByEl.delete(this)l
      }
      if (el) {
        return el;
      }
      return this.ownerDocument.getElementById(this.getAttribute('popuptoggletarget')) || null;
    },
  });
```

src/triggers.ts
```diff
@@ -0,0 +1,85 @@
+const closestElement: (selector: string, target: Element) => Element | null = (
```
keithamus, line 1: Could we perhaps make this refactoring in a standalone PR? That way each change becomes easier to review.

src/popover.ts
yinonov: can you please elaborate on

> We're missing the concept that these must reflect the attribute. We need to query the DOM if these aren't set with a weakmap.

keithamus: Yes apologies. 

The [spec stipulates that these properties _reflect_ their respective attributes](https://whatpr.org/html/8221/popover.html#dom-popovertoggletargetelement). So for example `popoverToggleTargetElement` must _reflect_ the `popovertoggletarget` attribute. [_Reflect_ is a concept in the DOM spec](https://whatpr.org/html/8221/common-dom-interfaces.html#reflect). In short, this means the property should derive its value from the attribute. An example of a reflected attribute is `classList` - it _reflects_ `class`, so if you change the `class` attribute, then `classList` changes in accordance, if you change `classList` and add or remove a class, then the `class` attribute is updated to match. 

So for the specifics of these properties, they must return the element with the ID that the respective attribute is set to. So for example if we had a `<button popoovertoggletarget="foo">`, getting the `popoverToggleTargetElement` property would return the element who's id is `foo`. If we _set_ the `popoverToggleTargetElement`, then the `popovertoggletarget` attribute value would change.

In my comment here: https://github.com/oddbird/popover-polyfill/issues/66#issuecomment-1405230146 I outlined the steps for this reflection. I'll quote that and add emphasis so you can see which steps they are:

> * `popoverToggleTargetElement`, `popoverShowTargetElement`, `popoverHideTargetElement` will all be properties that will be exposed on `HTMLElement`
> 
>     * We'll need a weakmap for each that stores has the key of the element (`this`) and the value that it can be set to (the target element).
> 
>     * The setter function should set the weakmap to the value given
> 
>     * The getter function will need to run through some steps ([reading from the spec](https://whatpr.org/html/8221/popover.html#popover-target-element)):
>       
>       * If the `el.disabled` return `null`
>       * If `el.form && isSubmitButton(el)` return `null`
>       * If the WeakMap value is not falsey
>         
>         * If value's `.popover` value is `'auto'` or `'manual'` returns the value
>         * Else return null
>       * **If the WeakMap value is empty...**
>         
>         * **Call `getAttribute()` with the respective attribute**
>         * **If the attribute is falsey, return `null`**
>         * **If the attribute is a string, call get `el = document.getElementById(attr)`**
>           
>           * If value's `.popover` value is `'auto'` or `'manual'` returns the value
>           * Else return null

There are some extra validation steps to ensure that we're always returning an appropriate popover element, but they are orthogonal to the concept of _reflect_.

There is an additional complication when it comes to reflecting back the value into the attribute - which is that `popoverTooggleTargetElement` can be set to an element which does not have an ID. In this case I _think_ the `popovertoggletarget` attribute should be removed. We can confirm this by checking Chrome's implementation and aligning to that for now, until a normative behaviour emerges or is specified.

yinonov: Thanks. I'm aware of the reflection and relation between attr and props. it's just that in this particular case properties reflect from attributes only (as in `fromView` mode) and not the other way around which puzzled me here. if property is set, it won't reflect to the attributes as it will result in something like `popuptoggletarget=[object object]`.
so we're talking unidirectional, correct?
thinking if `popuptoggletarget` is already set and then a property `popoverTooggleTargetElement ` is assigned, `popuptoggletarget` should be removed? (reading your last paragraph that's exactly what you address. got it)

I followed https://github.com/oddbird/popover-polyfill/issues/66#issuecomment-1405230146 precisely but then broke out to fit requirement in code, I can guide through the history if need.

in the implementation you suggest, if value is assigned to property, it takes precedence over the attribute.
```js
      if (el) {
        return el;
      }
      return this.ownerDocument.getElementById(this.getAttribute('popuptoggletarget')) || null;
```
meaning even if `idref` is set to `popuptoggletarget` attribute, it doesn't override the property value. if a value is set to that property it will return it even though a value is set for the attribute, thus will create confusion for authors. is that the expected behavior from spec?

keithamus: > it's just that in this particular case properties reflect from attributes only (as in fromView mode) and not the other way around which puzzled me here.

I believe reflection should always be in both directions. 

> if property is set, it won't reflect to the attributes as it will result in something like `popuptoggletarget=[object object]`.

Reflect has specific steps for unpacking types into representative types. If you take a look at https://whatpr.org/html/8221/common-dom-interfaces.html#reflect it says:

> If a reflecting IDL attribute attr has the type T?, where T is either [Element](https://dom.spec.whatwg.org/#interface-element) or an interface that inherits from [Element](https://dom.spec.whatwg.org/#interface-element), then:
> - Elements of the type this IDL attribute appears on have an explicitly set attr-element, which is a weak reference to an element or null. It is initially null.
> - Elements of the type this IDL attribute appears on have an attr-associated element. To compute the [attr-associated element](https://whatpr.org/html/8221/common-dom-interfaces.html#attr-associated-element) for such an element element:
>    1. If element's [explicitly set attr-element](https://whatpr.org/html/8221/common-dom-interfaces.html#explicitly-set-attr-element) is not null:
>        - If element's [explicitly set attr-element](https://whatpr.org/html/8221/common-dom-interfaces.html#explicitly-set-attr-element) is a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of any of element's [shadow-including ancestors](https://dom.spec.whatwg.org/#concept-shadow-including-ancestor), then return element's [explicitly set attr-element](https://whatpr.org/html/8221/common-dom-interfaces.html#explicitly-set-attr-element).
>        - Otherwise, return null.
>      2. Otherwise, if the content attribute is present on element, then return the first element candidate, in [tree order](https://dom.spec.whatwg.org/#concept-tree-order), that meets the following criteria:
            candidate's [root](https://dom.spec.whatwg.org/#concept-tree-root) is the same as element's [root](https://dom.spec.whatwg.org/#concept-tree-root),
            candidate's [ID](https://dom.spec.whatwg.org/#concept-id) is the value of the content attribute, and
            candidate [implements](https://webidl.spec.whatwg.org/#implements) T.
> If no such element exists, then return null.
>      3. Return null.

In other words: if `popoverToggleTargetElement` was explicitly set, we should return that. If not we should take the `popoverToggleTarget` attribute's contents, and use that to query the DOM and return the matching element, or return null.

> Chrome's implementation

[Here's the relevant lines from Chrome's impl for `set` of `popovertoggletargetelement`](https://source.chromium.org/chromium/chromium/src/+/main:out/chromeos-Debug/gen/third_party/blink/renderer/bindings/core/v8/v8_html_button_element.cc;l=480-515).

```c
void PopoverToggleTargetElementAttributeSetCallback(
    const v8::FunctionCallbackInfo<v8::Value>& info) {
  RUNTIME_CALL_TIMER_SCOPE_DISABLED_BY_DEFAULT(
      info.GetIsolate(),
      "Blink_HTMLButtonElement_popoverToggleTargetElement_Setter");
  BLINK_BINDINGS_TRACE_EVENT(
      "HTMLButtonElement.popoverToggleTargetElement.set");
  v8::Isolate* isolate = info.GetIsolate();
  const ExceptionState::ContextType exception_state_context_type =
      ExceptionContext::Context::kAttributeSet;
  const char* const class_like_name = "HTMLButtonElement";
  const char* const property_name = "popoverToggleTargetElement";
  ExceptionState exception_state(isolate, exception_state_context_type,
                                 class_like_name, property_name);
  if (UNLIKELY(info.Length() < 1)) {
    exception_state.ThrowTypeError(
        ExceptionMessages::NotEnoughArguments(1, info.Length()));
    return;
  }
  // [CEReactions]
  CEReactionsScope ce_reactions_scope;
  v8::Local<v8::Object> v8_receiver = info.This();
  HTMLButtonElement* blink_receiver =
      V8HTMLButtonElement::ToWrappableUnsafe(v8_receiver);
  v8::Local<v8::Value> v8_property_value = info[0];
  auto&& arg1_value = NativeValueTraits<IDLNullable<Element>>::NativeValue(
      isolate, v8_property_value, exception_state);
  if (UNLIKELY(exception_state.HadException())) {
    return;
  }
  blink_receiver->SetElementAttribute(html_names::kPopovertoggletargetAttr,
                                      arg1_value);
}
```

The `SetElementAttribute` call is what we're interested in. That's [this method](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/dom/element.cc;l=835-863;drc=3297fb6d70a02d8c4af8db2bc9e315053744442b):

```c
void Element::SetElementAttribute(const QualifiedName& name, Element* element) {
  ExplicitlySetAttrElementsMap* explicitly_set_attr_elements_map =
      GetDocument().GetExplicitlySetAttrElementsMap(this);
  // If the reflected element is explicitly null then we remove the content
  // attribute and the explicitly set attr-element.
  if (!element) {
    explicitly_set_attr_elements_map->erase(name);
    removeAttribute(name);
    return;
  }
  setAttribute(name, g_empty_atom);
  auto result = explicitly_set_attr_elements_map->insert(name, nullptr);
  if (result.is_new_entry) {
    result.stored_value->value =
        MakeGarbageCollected<HeapLinkedHashSet<WeakMember<Element>>>();
  } else {
    result.stored_value->value->clear();
  }
  result.stored_value->value->insert(element);
  if (isConnected()) {
    if (AXObjectCache* cache = GetDocument().ExistingAXObjectCache()) {
      cache->HandleAttributeChanged(name, this);
    }
  }
}
```

The big we care about is
```c
if (!element) {
  explicitly_set_attr_elements_map->erase(name);
  removeAttribute(name);
  return;
}
setAttribute(name, g_empty_atom);
```

`g_empty_atom` is an empty attribute value node, so in other words `""`.

From my reading of this, if `popoverToggleTargetElement` is set to `null`, then we should remove the `popovertoggletarget` attribute. If `popoverToggleTargetElement` is set to a valid element, then `popovertoggletarget` should be set to `""`.

yinonov: very helpful guidance 🙌 thank you!

yinonov:
should popoverTargetElement element check advance to [supported popover invokers](https://whatpr.org/html/8221/popover.html#attr-popover-target-supported-elements) check?
https://github.com/yinonov/popover-polyfill/blob/6c7176645aa396d242fd819795d872bb9b47a6e0/src/popover.ts#L213-L215

src/popover.ts
```diff
@@ -149,6 +163,90 @@ export function apply() {
     },
   });
 
+  const popoverTargetAttributesSupportedElementsSelector =
+    ':is(button, input[type="button"], input[type="submit"], input[type="image"], input[type="reset"])';
+
+  const definePopoverTargetElementProperty = (name: string) => {
+    const invokersMap = new WeakMap<Element, Element>();
+    Object.defineProperty(HTMLElement.prototype, `${name}Element`, {
```
keithamus, line 171: I think we probably want to define these on `HTMLButtonElement.prototype` and `HTMLInputElement.prototype`.

src/popover.ts
```diff
@@ -149,6 +163,90 @@ export function apply() {
...
+      },
+    });
+  };
+
+  definePopoverTargetElementProperty('popoverToggleTarget');
+  definePopoverTargetElementProperty('popoverShowTarget');
+  definePopoverTargetElementProperty('popoverHideTarget');
+
+  const handlePopoverTargetElementInvokation = (invoker: Element | null) => {
+    if (!(invoker instanceof HTMLButtonElement)) {
```
keithamus, line 227:
```suggestion
    if (!(invoker instanceof HTMLButtonElement) && !(invoker instanceof HTMLInputElement)) {
```

keithamus (commented):
This is looking close to shippable IMO. A couple of points though:

src/index.ts
```diff
@@ -7,6 +7,9 @@ declare global {
   }
   interface HTMLElement {
     popover: 'auto' | 'manual' | null;
+    popoverToggleTargetElement: HTMLElement | null;
+    popoverShowTargetElement: HTMLElement | null;
+    popoverHideTargetElement: HTMLElement | null;
     showPopover(): void;
     hidePopover(): void;
   }
```
yinonov, line 15: should we?
```suggestion
  interface HTMLElement {
    popover: 'auto' | 'manual' | null;
    showPopover(): void;
    hidePopover(): void;
  }
  interface HTMLButtonElement {
    popoverToggleTargetElement: HTMLElement | null;
    popoverShowTargetElement: HTMLElement | null;
    popoverHideTargetElement: HTMLElement | null;
  }
  interface HTMLInputElement {
    popoverToggleTargetElement: HTMLElement | null;
    popoverShowTargetElement: HTMLElement | null;
    popoverHideTargetElement: HTMLElement | null;
  }
```

keithamus: Yes please!

index.html
```diff
@@ -23,6 +23,7 @@ <h1>Popover Attribute Polyfill</h1>
       <div id="popover8" popover="auto">Popover 8 (auto)</div>
       <div id="popover9" popover="manual">Popover 9 (manual)</div>
       <div id="popover10" popover="manual">Popover 10 (manual)</div>
+      <div id="popover11" popover="auto">Popover 11 (auto)</div>
```
yinonov, line 26:
```suggestion
```

yinonov:
I think we're done @keithamus

src/popover.ts
```diff
@@ -149,70 +163,132 @@ export function apply() {
...
-    const root = target.getRootNode();
-    if (!(root instanceof ShadowRoot || root instanceof Document)) {
+  const popoverTargetAttributesSupportedElementsSelector =
+    ':is(button, input[type="button"], input[type="submit"], input[type="image"], input[type="reset"])';
+
+  const definePopoverTargetElementProperty = (name: string) => {
+    const invokersMap = new WeakMap<Element, Element>();
+    function obj(this: HTMLButtonElement | HTMLInputElement) {
       return;
     }
```
keithamus, line 173:
```suggestion
```

src/popover.ts
```diff
@@ -149,70 +163,132 @@ export function apply() {
...
 
-  const onClick = (event: Event) => {
-    const target = event.target;
-    if (!(target instanceof Element) || target?.shadowRoot) {
-      return;
-    }
-    const root = target.getRootNode();
-    if (!(root instanceof ShadowRoot || root instanceof Document)) {
+  const popoverTargetAttributesSupportedElementsSelector =
+    ':is(button, input[type="button"], input[type="submit"], input[type="image"], input[type="reset"])';
```
keithamus, line 167: Given this is a polyfill we might want to avoid selectors like `:is` which are fairly new. I don't think the `:is` adds anything here either.

```suggestion
    'button, input[type="button"], input[type="submit"], input[type="image"], input[type="reset"]';
```

src/popover.ts
```diff
@@ -149,70 +163,132 @@ export function apply() {
...
+        } else {
+          this.setAttribute(name.toLowerCase(), '');
+          invokersMap.set(this, targetElement);
+        }
+      },
+      get() {
+        // If node is not supported, then return null.
+        if (!this.matches(popoverTargetAttributesSupportedElementsSelector)) {
+          return null;
+        }
```
keithamus, line 191: We could drop `matches` here, and be a little more explicit:

```suggestion
        // If node is not supported, then return null.
        if (this.localName !== 'button' && this.localName !== 'input') {
          return null;
        }
        
        // If the node is not the correct input type, then return null.
        if (this.localName === 'input' && this.type !== 'reset' && this.type !== 'image' && this.type !== 'button') {
          return null;
        }
```

Matches is fine if you want to keep it, but considering we have the other logic inlined here, it would improve readability IMO

src/popover.ts
yinonov: how did that pass TS 🤔

src/popover.ts
yinonov: agree

yinonov: agree

keithamus (approved):
This is looking great! Excellent work! 

I'll leave @jgerigmeyer or @sanajaved7 to give a second review and merge, due to the size of the change. But this is :ok_hand: in my view.

src/popover.ts
mfreed7: Just wanted to check in here and say that this is the correct interpretation of the spec and behavior. Thanks for working on this polyfill!

src/popover.ts
```diff
@@ -149,6 +162,100 @@ export function apply() {
...
+      `${name}Element`,
+      invokerDescriptor,
+    );
+  };
+
+  definePopoverTargetElementProperty('popoverToggleTarget');
+  definePopoverTargetElementProperty('popoverShowTarget');
+  definePopoverTargetElementProperty('popoverHideTarget');
+
+  const handlePopoverTargetElementInvokation = (invoker: Element | null) => {
```
jgerigmeyer, line 232:
```suggestion
  const handlePopoverTargetElementInvocation = (invoker: Element | null) => {
```

src/popover.ts
```diff
@@ -158,61 +265,18 @@ export function apply() {
...
-        if (visibleElements.has(effectedPopover)) {
-          effectedPopover.hidePopover();
-        } else {
-          effectedPopover.showPopover();
-        }
-      }
-    }
-
-    // Dismiss open Popovers
+    const popoverTargetElement = handlePopoverTargetElementInvokation(invoker);
```
jgerigmeyer, line 271:
```suggestion
    const popoverTargetElement = handlePopoverTargetElementInvocation(invoker);
```

jgerigmeyer (approved):
One minor spelling suggestion, but this looks great! Thanks for the contribution. 🎉

jgerigmeyer:
Published in v0.0.9 🚀

PR #77, opened 2023-03-02
alter aria-expanded attributes for popover invokers

keithamus:
Native popover in Chrome exposes "expanded" accessibility info for popover invokers, which allows assistive technologies to know that the button will expand some kind of dialog or menu.

This PR adds `aria-expanded=false` on all popover invoker buttons that target an `auto` popover, during page load or insertion into the DOM, and when a popover becomes visible, it will toggle `aria-expanded=true` for all invokers of a visible popover.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@
 ## UNRELEASED
 
 - 🏠 INTERNAL: Upgrade dependencies
+- 🚀 NEW: Add support for `aria-expanded` on invokers --
+  [#77](https://github.com/oddbird/popover-polyfill/pull/77)
 
 ## 0.0.9: 2023-02-03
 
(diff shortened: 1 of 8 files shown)


jgerigmeyer (approved):
Looks good! Two thoughts:

1. Would it make sense to expand the README documentation to mention e.g. `aria-expanded="false"`?
2. Could you update the CHANGELOG to add this change?

keithamus:
Actually this might need rethinking a little, because it interferes with the native behaviour (as aria-expanded is then fixed to `"false"` in supported browsers)

keithamus:
This has been updated so that it sets `aria-expanded` on all button/input elements that are invokers for a `popover=auto`.

jgerigmeyer (approved):

keithamus:
@jgerigmeyer shall I add a version bump commit?

jgerigmeyer:
> @jgerigmeyer shall I add a version bump commit?

Go ahead and merge -- I'll do a version bump and release this afternoon 👍

PR #82, opened 2023-03-14
Add escape to dismiss

keithamus:
This PR adds support for the `Escape` key causing all auto popovers to be dismissed.

## Steps to test/reproduce
- Enact the `Click to show Popover 3` button
- Observe popover 3 is open
- Press Escape
- Observe popover 3 is now closed.

## Show me

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -265,8 +265,16 @@ export function apply() {
     );
   };
 
+  const onKeydown = (event: Event) => {
+    const key = (event as KeyboardEvent).key;
+    if (key === 'Escape' || key === 'Esc') {
+      hideOpenAutoPopovers();
+    }
+  };
+
   const addOnClickEventListener = (root: Document | ShadowRoot) => {
     root.addEventListener('click', onClick);
+    root.addEventListener('keydown', onKeydown);
   };
 
   observePopoversMutations(document);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- 🚀 NEW: Add support for Escape to dismiss auto popovers --
+  [#82](https://github.com/oddbird/popover-polyfill/pull/82)
 - 🚀 NEW: Showing an `auto` popover closes other `auto` popovers --
   [#83](https://github.com/oddbird/popover-polyfill/pull/83)
 - 🚀 NEW: Add support for focus restoration on popover close --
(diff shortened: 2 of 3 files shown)


src/popover.ts
```diff
@@ -284,22 +300,19 @@ export function apply() {
...
-      }
-    }
+    dismissAutoPopovers(event, popoverTargetElement);
   };
 
-  const addOnClickEventListener = (
-    (callback: (event: Event) => void) => (root: Document | ShadowRoot) => {
-      root.addEventListener('click', callback);
+  const onKeydown = (event: Event) => {
+    if ((event as KeyboardEvent).key === 'Escape') {
```
jgerigmeyer, line 307: I'm in the habit of also checking for `'Esc'` in cases like this, but that's probably outdated enough by now that we don't need to worry about it?

src/popover.ts
```diff
@@ -273,6 +273,22 @@ export function apply() {
...
+  function dismissAutoPopovers(
+    event: Event,
+    popoverTargetElement: HTMLElement | null = null,
+  ) {
+    const target = event.target;
+    for (const popover of popovers) {
+      if (
+        popover.matches('[popover="" i].\\:open, [popover=auto i].\\:open') &&
+        popover !== popoverTargetElement &&
+        !queryAncestorAll(target as Element, '[popover]').includes(popover)
```
jgerigmeyer, line 285: In the case of Popover 3, where the internal link receives auto-focus, is it intentional that `Esc` does not dismiss the popover when focus is on the link? It only dismisses once focus is first moved to another element, not within the popover.

I haven't dug into the spec to see if this is to be expected -- I just found it a bit surprising while testing this out.

src/popover.ts
keithamus: I think `'Esc'` will only be output in IE/EdgeHTML browsers, which I don't think will run this code for other reasons. Having said that I don't mind adding it in with a view to later support those browsers if that's a desire.

src/popover.ts
keithamus: This sounds like an issue for sure! I didn't notice this but I'll try to reproduce!

keithamus: I managed to reproduce but also discovered another bug, which is fixed in #83 - if we could merge that one first then I can rebase this one on top to bring in some of the refactored code here.

keithamus:
@jgerigmeyer this one has been rebased and is, IMO, good to go.

PR #84, opened 2023-03-16
Patch qsa matches closest to accept `:popover-open`

keithamus:
A potentially controversial change.

By default running something like `el.matches(':open')` on a browser without native Popover will result in an exception:

Chrome:
```
DOMException: Failed to execute 'matches' on 'Element': ':open' is not a valid selector.
```

Firefox:
```
DOMException: Element.matches: ':open' is not a valid selector
```

This means users of this polyfill need to work around this; either by using a `try`/`catch`, or by doing feature detection up front, e.g.:

```js
let openSelector = ':open'
try {
  document.querySelector(openSelector)
} catch {
  // use polyfilled selector instead
  openSelector = '.\\:open'
}
```

This is a little cumbersome and has to be done for every selector you intend to use, and given we're patching prototype methods anyway, perhaps it could be useful for us to patch the CSS selector functions (I believe I have them all, `querySelector`, `querySelectorAll`, `matches`, `closest`).

This change patches all of those functions taking the selector string given and replacing instances of `:open` with `.\\:open` and `:closed` with `:not(.\\:open)`. This allows us to use the `:open`/`:closed` selector and have it work with the polyfill.

## Steps to test/reproduce

Tests have been added to ensure qsa/matches/closed work but it doesn't do anything close to exhaustive input testing. I've manually tested the replacer function with various inputs. We could add some unit tests to ensure its robustness, but here's how I validated it:

```js
const nonEscapedOpenSelector = /(^|[^\\]):open\b/g;
const nonEscapedClosedSelector = /(^|[^\\]):closed\b/g;
  function rewriteSelector(selector) {
    return selector
      .replace(nonEscapedOpenSelector, '$1.\\:open')
      .replace(nonEscapedClosedSelector, '$1:not(.\\:open)');
  }
console.assert(rewriteSelector(':open') == '.\\:open')
console.assert(rewriteSelector('.\\:open') == '.\\:open')
console.assert(rewriteSelector('[bar]:open') == '[bar].\\:open')
console.assert(rewriteSelector('[bar]:opened') == '[bar]:opened')
console.assert(rewriteSelector('[bar]:open.\\:open:open') == '[bar].\\:open.\\:open.\\:open')
console.assert(rewriteSelector(':closed') == ':not(.\\:open)')
console.assert(rewriteSelector('.\\:closed') == '.\\:closed')
```

## Show me

### DIFF
--- a/src/popover.css
+++ b/src/popover.css
@@ -13,6 +13,9 @@
 }
 
 /* stylelint-disable selector-class-pattern */
+
+/* This older `:open` pseudo selector is deprecated and support will be removed
+in a later release. */
 @supports not selector([popover]:open) {
   [popover]:not(.\:popover-open, dialog[open]) {
     display: none;
--- a/src/popover.css
+++ b/src/popover.css
@@ -14,12 +14,22 @@
 
 /* stylelint-disable selector-class-pattern */
 @supports not selector([popover]:open) {
-  [popover]:not(.\:open) {
+  [popover]:not(.\:popover-open, dialog[open]) {
     display: none;
+  }
+
+  [anchor].\:popover-open {
+    inset: auto;
+  }
+}
+
+@supports not selector([popover]:popover-open) {
+  [popover]:not(.\:popover-open, dialog[open]) {
+    display: none;
+  }
+
+  [anchor].\:popover-open {
+    inset: auto;
   }
 }
 /* stylelint-enable selector-class-pattern */
-
-[popover][anchor] {
-  inset: auto;
-}
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -16,8 +16,50 @@ export function isSupported() {
   );
 }
 
+function patchSelectorFn<K extends string>(
+  object: Record<PropertyKey & K, unknown>,
+  name: K,
+  mapper: (selector: string) => string,
+) {
+  const original = object[name] as (selectors: string) => NodeList;
+  Object.defineProperty(object, name, {
+    value(selector: string) {
+      return original.call(this, mapper(selector));
+    },
+  });
+}
+
+const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
+
 export function apply() {
   window.ToggleEvent = window.ToggleEvent || ToggleEvent;
+
+  function rewriteSelector(selector: string) {
+    if (selector.includes(':popover-open')) {
+      selector = selector.replace(
+        nonEscapedPopoverSelector,
+        '$1.\\:popover-open',
+      );
+    }
+    return selector;
+  }
+
+  patchSelectorFn(Document.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(Document.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'matches', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'closest', rewriteSelector);
+  patchSelectorFn(
+    DocumentFragment.prototype,
+    'querySelectorAll',
+    rewriteSelector,
+  );
+  patchSelectorFn(
+    DocumentFragment.prototype,
+    'querySelectorAll',
+    rewriteSelector,
+  );
 
   Object.defineProperties(HTMLElement.prototype, {
     popover: {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Popover Attribute Polyfill Changelog
 
+## 0.2.0: 2023-03-24
+
+- 🚀 NEW: Support for `:popover-open` psuedo selector
+  [#84](https://github.com/oddbird/popover-polyfill/pull/84)
+
 ## 0.1.1: 2023-03-24
 
 - 🐛 BUGFIX: Fix regression when targets have nested elements --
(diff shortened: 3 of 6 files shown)


src/popover.ts
```diff
@@ -24,10 +24,49 @@ function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
...
+  }
+
+  patchSelectorFn(HTMLDocument.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(HTMLDocument.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'matches', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'closest', rewriteSelector);
+  patchSelectorFn(
+    DocumentFragment.prototype,
```
keithamus, line 60: `ShadowRoot` inherits `DocumentFragment`, and so patching `DocumentFragment` also patches `ShadowRoot`.

src/popover.ts
```diff
@@ -24,10 +24,49 @@ function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
...
+      .replace(nonEscapedOpenSelector, '$1.\\:open')
+      .replace(nonEscapedClosedSelector, '$1:not(.\\:open)');
+  }
+
+  patchSelectorFn(HTMLDocument.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(HTMLDocument.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'querySelectorAll', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'matches', rewriteSelector);
+  patchSelectorFn(Element.prototype, 'closest', rewriteSelector);
```
keithamus, line 58: `HTMLDocument` nor `DocumentFragment` have `matches` nor `closest`. I think I have all functions that take CSS selectors here. Let me know if I missed one!

src/popover.ts
```diff
@@ -24,10 +24,49 @@ function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
...
   let lastFocusedElement: HTMLElement | null = null;
 
+  function rewriteSelector(selector: string) {
+    return selector
+      .replace(nonEscapedOpenSelector, '$1.\\:open')
+      .replace(nonEscapedClosedSelector, '$1:not(.\\:open)');
+  }
+
+  patchSelectorFn(HTMLDocument.prototype, 'querySelector', rewriteSelector);
+  patchSelectorFn(HTMLDocument.prototype, 'querySelectorAll', rewriteSelector);
```
keithamus, line 54: I realised this should be `Document`

```suggestion
  patchSelectorFn(Document.prototype, 'querySelector', rewriteSelector);
  patchSelectorFn(Document.prototype, 'querySelectorAll', rewriteSelector);
```

src/popover.ts
```diff
@@ -24,10 +24,49 @@ function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
...
+
+const nonEscapedOpenSelector = /(^|[^\\]):open\b/g;
+const nonEscapedClosedSelector = /(^|[^\\]):closed\b/g;
+
 export function apply() {
   const visibleElements = new WeakSet<HTMLElement>();
   let lastFocusedElement: HTMLElement | null = null;
 
+  function rewriteSelector(selector: string) {
+    return selector
```
keithamus, line 48: Adding an early return here will speed the unhappy path up a bit:

```suggestion
    if (!selector.includes(':open')) {
      return selector;
    }
    return selector
```

src/popover.ts
```diff
@@ -24,10 +24,52 @@ function patchAttachShadow(callback: (shadowRoot: ShadowRoot) => void) {
...
+
+const nonEscapedOpenSelector = /(^|[^\\]):open\b/g;
+const nonEscapedClosedSelector = /(^|[^\\]):closed\b/g;
+
 export function apply() {
   const visibleElements = new WeakSet<HTMLElement>();
   let lastFocusedElement: HTMLElement | null = null;
 
+  function rewriteSelector(selector: string) {
+    if (!selector.includes(':open')) {
```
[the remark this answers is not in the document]
jgerigmeyer: I think that's causing the failing tests too...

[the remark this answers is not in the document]
keithamus: Yes! :man_facepalming:

jgerigmeyer (approved):

keithamus:
Looks like https://github.com/w3c/csswg-drafts/issues/8637 is moving forward with `:popover`. We should probably wait for https://github.com/whatwg/html/pull/9077 to land before we merge this.

keithamus:
Upstream PRs have landed. `:popover-open` is the new selector. We should consider re-reviewing this PR now.

src/popover.css
```diff
@@ -14,12 +14,22 @@
 
 /* stylelint-disable selector-class-pattern */
 @supports not selector([popover]:open) {
```
jgerigmeyer, line 16: Is it important to retain this older `:open` selector as well? I assume just temporarily, then we might remove it?

jgerigmeyer (approved):

jgerigmeyer:
Add note to changelog?

src/popover.css
keithamus: Yeah exactly

jgerigmeyer (approved):

PR #86, opened 2023-03-22
Refactor to be much closer to spec

keithamus:
This is a rather large change, I do apologise for the size, it perhaps got away with me. It may be difficult to break down smaller.

Now the spec is fully defined, and most changes have settled, we can adhere to it more closely. This PR changes the code to more closely follow the set of algorithms defined in the spec, and consequently it makes some breaking changes to align with the proper spec, rather than the PR:

- `popovertoggletarget`, `popovershowtarget` and `popoverhidetarget` are no longer valid. Instead we have `popovertarget` which defines the target, and `popovertargetaction` which defines the action (defaulting to toggle).
- Invalid `popover` attribute values now default to `manual`, meaning `popover=invalid` is, in fact, a valid popover.
- The old `BeforeToggleEvent` has been replaced with `ToggleEvent` which has a type of `beforetoggle` or `toggle`, and the old `currentState` is now `oldState`. `newState` remains the same.

## Steps to test/reproduce
The tests validate the work, but a variety of behaviours have subtly changed, so some exploratory manual testing is recommended. I've done a bunch of manual testing and I'm satisfied.

## Show me


src/popover.ts
```diff
@@ -229,57 +160,28 @@ export function apply() {
...
-    );
   };
 
   const onKeydown = (event: Event) => {
     const key = (event as KeyboardEvent).key;
-    if (key === 'Escape' || key === 'Esc') {
-      hideOpenAutoPopovers();
+    const target = event.target as Element;
+    if (target && (key === 'Escape' || key === 'Esc')) {
+      hideAllPopoversUntil(target.ownerDocument, false, false);
```
jgerigmeyer, line 174: Is it intentional that this does not return focus to the previously-focused element?

keithamus: It's not. This should be `true`.

```suggestion
      hideAllPopoversUntil(target.ownerDocument, true, false);
```

keithamus:
I've added some refinements and fixed some edge cases. This now also works on ShadowRoots without having to patch attachShadow by using `composedPath()`.

src/popover-helpers.ts
```diff
@@ -1,44 +1,359 @@
...
-export function closestShadowPenetrating(
+// https://html.spec.whatwg.org/#hide-all-popovers-until
+export function hideAllPopoversUntil(
+  endpoint: Element | Document,
+  focusPreviousElement: boolean,
+  fireEvents: boolean,
+) {
+  const document = endpoint.ownerDocument || endpoint;
+  if (endpoint instanceof Document) {
+    return closeAllOpenPopovers(document);
```
jgerigmeyer, line 310:
```suggestion
    return closeAllOpenPopovers(document, focusPreviousElement, fireEvents);
```

src/popover-helpers.ts
```diff
@@ -1,44 +1,359 @@
...
+  for (const popover of autoPopoverList.get(document) || []) {
+    if (popover === endpoint) {
+      foundEndpoint = true;
+    } else if (foundEndpoint) {
+      lastToHide = popover;
+      break;
+    }
+  }
+  if (!foundEndpoint) {
+    return closeAllOpenPopovers(document);
```
jgerigmeyer, line 323:
```suggestion
    return closeAllOpenPopovers(document, focusPreviousElement, fireEvents);
```

jgerigmeyer (approved):

keithamus:
@jgerigmeyer I'm quite happy with this one now. Shall I go ahead and merge?

jgerigmeyer:
> @jgerigmeyer I'm quite happy with this one now. Shall I go ahead and merge?

Works for me! I'm tempted to publish this refactor as 0.1.0 -- what do you think?

keithamus:
As for `0.1.0`, that sounds good to me. As long as we're pre 1.0 then the implied contract is caveat emptor, regardless of minor/patch :smile:

PR #88, opened 2023-03-24
rewrite nearestInclusive.* algos to be closer to the spec

keithamus:
In a misinterpretation of the spec in #86 a regression was introduced whereby the invoker can cause toggles to be no-ops. The reason is largely to do with the `nearestInclusiveTargetPopoverForInvoker` which was not written close to the spec, instead it was using `nearestInclusivePopover(target.popoverTarget)` which is not appropriate if the target contained nested elements like spans, or a composed event came through a shadow boundary.

This fixes that by rewriting these algos to use the spec methodology.

## Steps to test/reproduce
Tests should cover these scenarios.

Try clicking the `Click to toggle shadowed Popover (auto)` though, which should toggle, and you might find once it's visible it cannot be closed with the same button.

Try clicking the `Popover 1` button - click the button and it should work, click the _text_ (the span) and you'll see it won't.

## Show me

### DIFF
--- a/index.html
+++ b/index.html
@@ -73,7 +73,7 @@
 
     <div id="buttons">
       <button popovertarget="popover1" aria-expanded="false">
-        Click to toggle Popover 1
+        <span>Click to toggle Popover 1</span>
       </button>
       <button popovertarget="popover2">Click to toggle Popover 2</button>
       <button popovertargetaction="show" popovertarget="popover3">
--- a/index.html
+++ b/index.html
@@ -50,7 +50,6 @@
       <div id="host"></div>
 
       <script type="module">
-        import sheet from './dist/popover.css' assert { type: 'css' };
         const host = document.getElementById('host');
         const shadowRoot = host.attachShadow({ mode: 'open' });
         shadowRoot.innerHTML = `<button popovertarget="shadowedPopover">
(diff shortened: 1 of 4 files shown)


jgerigmeyer (approved):

jgerigmeyer (commented):
I'm not able to reproduce this issue on `main`, or any difference with this PR, but 🤷‍♂️

keithamus:
Actually I investigated the problem a little further and I've come up with the solution and a more accurate finding of the problem.

Following up with a rewrite of this shortly.

index.html
```diff
@@ -63,15 +62,18 @@ <h1>Popover Attribute Polyfill</h1>
                 <div>
                   <div id="shadowedNestedPopover" popover="auto">Shadowed Nested Popover (auto)</div>
                 </div>`;
-        shadowRoot.adoptedStyleSheets = [sheet];
+        const link = document.createElement('link');
+        link.rel = 'stylesheet';
+        link.href = './dist/popover.css';
+        shadowRoot.prepend(link);
```
jgerigmeyer, line 68: 👏

index.html
```diff
@@ -63,15 +62,18 @@ <h1>Popover Attribute Polyfill</h1>
...
         document.getElementById('crossTreeToggle').popoverTargetElement =
           shadowRoot.getElementById('shadowedPopover');
       </script>
     </div>
 
     <div id="buttons">
       <button popovertarget="popover1" aria-expanded="false">
-        Click to toggle Popover 1
+        <span>Click to toggle Popover 1</span>
```
jgerigmeyer, line 76: With this change on `main` I am able to reproduce the bug, and confirm that it is fixed in this PR.

jgerigmeyer (approved):
🚀

PR #99, opened 2023-05-17
remove disconnected autoPopovers during topMostPopoverAncestor algo

keithamus:
Prior to this change, a popover removed from the document would still be in the autoPopovers set, and subsequently `topMostPopverAncestor` would return disconnected elements, causing `closeAllPopovers` to hit an infinite loop:

1. `hidePopover` is called
2. `closeAllPopovers` is called
3. `topMostPopoverAncestor` is called and returns disconnected element.
4. `hidePopover(topMost)` is called but returns false, as the element is disconnected.
5. `topMostPopoverAncestor` is called, but it returns the same element. Repeat step 4, infinite loop.

This commit alters `topMostAutoPopover` to check if the element is disconnected, and remove it from the set if so, only returning the first connected element.

1. `hidePopover` is called
2. `closeAllPopovers` is called
3. `topMostPopoverAncestor` is called, the first element is disconnected, so it skips and returns the first connected element.
4. `hidePopover(topMost)` is called and returns true.
5. No infinite loop.

## Steps to test/reproduce
I've added a new test file "edge-cases.spec". This includes a test which fails on the current build (times out due to the browser hanging), but passes with the changes.

## Show me

Special thanks to @hmaurer for giving me an easy repro to this one!

### DIFF
--- a/src/popover-helpers.ts
+++ b/src/popover-helpers.ts
@@ -78,6 +78,7 @@ function topMostAutoPopover(document: Document): HTMLElement | null {
   const documentPopovers = autoPopoverList.get(document);
   for (const popover of documentPopovers || []) {
     if (!popover.isConnected) {
+      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
       documentPopovers!.delete(popover);
     } else {
       return popover;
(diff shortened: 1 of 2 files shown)


keithamus:
@jgerigmeyer could we get this one out as a `0.2.1` release please?

jgerigmeyer (approved):
🎉

PR #103, opened 2023-06-01
tweak styles & guard some code in older browsers

keithamus:
This PR changes the CSS and JS in some ways that support various combinations of esoteric browsers+flags. It fixes the following:

- Chrome 113 with Experimental Web Platform Features enabled. This browser supports native popover but with the `:open` psuedo selector (not `:popover-open`).
- Safari Technical Preview 166-171. These browsers support `popover` but either with `:open`, or both `:open` _and_ `:popover-open`.
- Older versions of Chromium such as 87, which includes alts like Qutebrowser, and older versions of Firefox such as Firefox 91, which includes alts like Seamonkey, Watefox Classic. These browsers don't support `@supports selector()`, `width: fit-content`, `inset`, nor do they support `node.getRootNode()`, `ShadowRoot`, or `HTMLDialogElement`. Luckily these are all fairly straightforward things to smooth over.

As a result of the combinatorial complexity of trying to support `:popover-open` and `:open` as well as support browsers which themselves don't support `@supports selector()`, the easiest solution is to avoid `@supports not(selector(...))` and instead use `display: revet` for the browsers that support `@supports selector()` and some combination of `:popover-open`/`:open`.

This should raise the support level to old browsers such as Firefox 91, Chrome ~80

I haven't done any testing in old Trident browsers like old Edge/IE.

## Steps to test/reproduce
Compare this branch to main in the following browsers:

 - Safari Technical Preview 166
 - Chrome 113 (with Experimental Web Platform Features enabled)
 - Seamonkey 2.53.16 (Firefox 91)
 - Waterfox Classic 2022.11
 - Qutebrowser 2.5.4 (Chromium 87)

## Show me

### DIFF
--- a/index.html
+++ b/index.html
@@ -47,6 +47,7 @@
           </ul>
         </div>
       </div>
+      <dialog popover>I'm a dialog!</dialog>
       <div id="host"></div>
 
       <script type="module">
--- a/src/popover-helpers.ts
+++ b/src/popover-helpers.ts
@@ -1,5 +1,11 @@
 import { queuePopoverToggleEventTask, ToggleEvent } from './events.js';
 
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const ShadowRoot = globalThis.ShadowRoot || function () {};
+
+// eslint-disable-next-line @typescript-eslint/no-empty-function
+const HTMLDialogElement = globalThis.HTMLDialogElement || function () {};
+
 const topLayerElements = new WeakMap<Document, Set<HTMLElement>>();
 const autoPopoverList = new WeakMap<Document, Set<HTMLElement>>();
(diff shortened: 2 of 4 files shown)


index.html
```diff
@@ -47,6 +47,7 @@ <h1>Popover Attribute Polyfill</h1>
           </ul>
         </div>
       </div>
+      <dialog popover>I'm a dialog!</dialog>
```
jgerigmeyer, line 50: Is this used/useful somewhere?

jgerigmeyer (approved):

index.html
keithamus: Yes I used this to manually test dialogs to ensure they function as expected, given https://github.com/oddbird/popover-polyfill/pull/103/files#diff-db819e90c0d1e17da5c46e2921746208de43ceae1c1843143c47e196cef30068R23-R25

PR #105, opened 2023-06-02
fix: querySelector APIs support null/undefined (closes #104)

### LINKED ISSUE #104: Polyfill breaks other libraries that pass null/undefined to document.querySelector
To demonstrate this, you can see this in action if you go to the [demo](https://popover-polyfill.netlify.app/) in a browser that doesn't natively implement popover, and in the console execute:

```js
document.querySelector(null);
```

This causes some interop issues with other 3rd party libraries.

## Expected behavior
`document.querySelector` does not throw an error, and returns null.

## Actual behavior
Error thrown: Uncaught TypeError: can't access property "includes", selector is null

Firefox:  
image

p-ob:
To demonstrate this, you can see this in action if you go to the [demo](https://popover-polyfill.netlify.app/) in a browser that doesn't natively implement popover, and in the console execute:

```js
document.querySelector(null);
```

This causes some interop issues with other 3rd party libraries.

## Expected behavior
`document.querySelector` does not throw an error, and returns null.

## Actual behavior
Error thrown: Uncaught TypeError: can't access property "includes", selector is null

Firefox:  
image

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -34,8 +34,8 @@ const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
 export function apply() {
   window.ToggleEvent = window.ToggleEvent || ToggleEvent;
 
-  function rewriteSelector(selector: string) {
-    if (selector.includes(':popover-open')) {
+  function rewriteSelector(selector: string | null | undefined) {
+    if (selector?.includes(':popover-open')) {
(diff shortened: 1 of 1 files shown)


keithamus (approved):
I wonder who's calling `querySelector` with `null`, but, defence in depth :joy:

jgerigmeyer (approved):

PR #108, opened 2023-06-22
Fix link attribute on README

markcellus:
Changes `src` to `href` so that it's valid syntax.

## Steps to test/reproduce
Look at the README? :)


keithamus (approved):
Oops :blush: good catch!

PR #126, opened 2023-09-14
ensure `:open` styles are applied for browsers that claim support

keithamus:
## Description

Gnome Web applies blocks inside `@supports (:open)` despite supporting `popover` or having UA styles for `popover`.

This applies the default UA style of `display:none` for a hidden popover, to avoid situations with Gnome Web where popovers remain always open.

## Steps to test/reproduce

Visit the demo site in Gnome Web.

/cc @webreflection if you'd kindly test this in Gnome Web for me that would be very helpful!

### DIFF
--- a/src/popover.css
+++ b/src/popover.css
@@ -35,6 +35,10 @@ in a later release. */
     display: revert;
   }
 
+  [popover]:not(:open) {
+    display: none;
+  }
+
   /* stylelint-disable selector-pseudo-class-no-unknown */
   [anchor]:is(:open) {
     inset: auto;
@@ -45,6 +49,10 @@ in a later release. */
 @supports selector([popover]:popover-open) {
   [popover]:not(.\:popover-open, dialog[open]) {
     display: revert;
+  }
+
+  [popover]:not(:popover-open) {
+    display: none;
   }
 
   /* stylelint-disable selector-pseudo-class-no-unknown */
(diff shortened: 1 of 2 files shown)


jgerigmeyer (approved):

keithamus:
Looks like this does not fix the underlying issues. I'll keep investigating.

src/popover.css
```diff
@@ -28,32 +27,35 @@
   inset: auto;
 }
 
-/* This older `:open` pseudo selector is deprecated and support will be removed
-in a later release. */
+/* This older `:open` pseudo selector is deprecated
+   and support will be removed in a later release. */
 @supports selector([popover]:open) {
+  [popover]:not(:open) {
```
WebReflection, line 33: `@supports selector([popover]:not(:open))` ??? you test one selector then you apply a more complex one :thinking:

src/popover.css
```diff
@@ -28,32 +27,35 @@
...
 
-  /* stylelint-disable selector-pseudo-class-no-unknown */
   [anchor]:is(:open) {
     inset: auto;
   }
-  /* stylelint-enable selector-pseudo-class-no-unknown */
 }
 
 @supports selector([popover]:popover-open) {
+  [popover]:not(:popover-open) {
```
WebReflection, line 47: `@supports selector([popover]:not(:popover-open))` ??? you test one selector then you apply a more complex one :thinking:

WebReflection:
why is this closed? and btw, I think this is related https://bugs.webkit.org/show_bug.cgi?id=258290

PR #128, opened 2023-09-18
drop :open styles

keithamus:
## Description

Older versions of Chrome with experimental web platform features turned on (Chrome 113) supported the `:open` pseudo class to determine if a popover was open or not. We have support code for this in the form of CSS that checks `@supports ([popover]:open)`.

The problem is that now causes issues with certain versions of WebKit, which claim to support `:open` and pass the CSS check, only to _unapply_ the popover styles but not actually support `popover`. In other words, there are versions of WebKit - such as Gnome Web - for which this breaks.

Given the popularity of Gnome Web (a fairly small % of total market share) compared with Chrome 113 with Experimental Web Platform Features enabled (an even smaller %), it makes sense to _remove_ this CSS in order to support browsers like Gnome Web.

## Steps to test/reproduce
Visit the demo page in Gnome Web. Everything is broken.
Visit this PR demo page in Gnome Web. Everything should work.

/cc @WebReflection

## Show me

### DIFF
--- a/src/popover.css
+++ b/src/popover.css
@@ -28,20 +28,6 @@
   inset: auto;
 }
 
-/* This older `:open` pseudo selector is deprecated and support will be removed
-in a later release. */
-@supports selector([popover]:open) {
-  [popover]:not(.\:popover-open, dialog[open]) {
-    display: revert;
-  }
-
-  /* stylelint-disable selector-pseudo-class-no-unknown */
-  [anchor]:is(:open) {
-    inset: auto;
-  }
-  /* stylelint-enable selector-pseudo-class-no-unknown */
-}
-
 @supports selector([popover]:popover-open) {
   [popover]:not(.\:popover-open, dialog[open]) {
     display: revert;


WebReflection:
Thank you **very much** for this and I can confirm everything works as expected now in Web: LGTM

If worth anything, here my thoughts:

```css
/* This older `:open` pseudo selector is deprecated and support will be removed
in a later release. */
```

I think that comment says it all, hopefully the lesson learned here is to drop ASAP any experimental hack for forever green browsers (such as Chrome/ium).

On the other hand, a `@support([any-attribute]:pseudo)` seems to be a way too generic check to me, and I wonder if there's any way that `@support` could be improved to know if such support is experimental ... is this maybe something to discuss with the *CSS* working group? an `:experimental` selector that exists only if a feature is, effectively, experimental?

This would allow more fine-tuning experiments targeting earlier versions (as it was in this case) without breaking browsers that implemented the "not experimental anymore" behavior.

/cc @tabatkins for thoughts about this (in case is not possible already)

jgerigmeyer (approved):

WebReflection:
@keithamus apologies for the late ping but ... any ETA when this will be deployed on GitHub ??? I still see the popover and it keeps being annoying :sweat_smile: thanks!

keithamus:
It’ll go out likely this week.

PR #150, opened 2023-11-24
Check slotted content for autofocus delegate

### LINKED ISSUE #149: `focusDelegate` method doesn't handle custom elements where `[autofocus]` is in the light DOM (slotted)
The `focusDelegate` method (https://github.com/oddbird/popover-polyfill/blob/66d6615eb0275784fb5492c34ba373ccb6b096c0/src/popover-helpers.ts#L188) checks if the `focusTarget` has a `shadowRoot` and if `delegatesFocus` is enabled.

If it does, it looks for `[autofocus]` in the shadow root and returns it if found.

It doesn't take the scenario into account where the `[autofocus]` element is in the light DOM and *slotted* into the shadow root. `focusDelegate` will return the `shadowRoot` itself in that case. 

When it is back in the `popoverFocusingSteps` method it tries to call `focus()` on the return value, but since `shadowRoot` does not implement that method, it causes an error (`Uncaught TypeError: _a.focus is not a function` in FF).

jpzwarte:
Fixes #149

### DIFF
--- a/src/popover-helpers.ts
+++ b/src/popover-helpers.ts
@@ -196,18 +196,22 @@ function focusDelegate(focusTarget: HTMLElement) {
   if (whereToLook.shadowRoot) {
     whereToLook = whereToLook.shadowRoot;
   }
-  let autoFocusDelegate: Element | null | undefined =
-    whereToLook.querySelector('[autofocus]');
+  let autoFocusDelegate = whereToLook.querySelector('[autofocus]');
   if (autoFocusDelegate) {
     return autoFocusDelegate;
   } else {
     const slots = whereToLook.querySelectorAll('slot');
     for (const slot of slots) {
-      autoFocusDelegate = slot
-        .assignedElements({ flatten: true })
-        .find((el) => el.hasAttribute('autofocus'));
-      if (autoFocusDelegate) {
-        return autoFocusDelegate;
+      const assignedElements = slot.assignedElements({ flatten: true });
+      for (const el of assignedElements) {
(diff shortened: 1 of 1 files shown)


keithamus (approved):
The change LGTM. A minor nit is that the `focusDelegate` function comes directly from the spec, but it has now diverged from the spec text quite significantly. We should probably add some commentary that this is no longer spec compliant (even though the current function is desirable). Perhaps a follow up can be made here?

jpzwarte:
> The change LGTM. A minor nit is that the `focusDelegate` function comes directly from the spec, but it has now diverged from the spec text quite significantly. We should probably add some commentary that this is no longer spec compliant (even though the current function is desirable). Perhaps a follow up can be made here?

I haven't read the spec, but as far as this PR is concerned: in Chrome & Safari, a slotted `[autofocus]` element has the same behavior as this PR.

I assume you will merge the PR since I can't do it?

keithamus:
> I haven't read the spec, but as far as this PR is concerned: in Chrome & Safari, a slotted `[autofocus]` element has the same behavior as this PR.

Yes I agree, getting the behaviour right is important. The concern was around the readability of this polyfill as all methods are mapped 1-1 to the spec, as much as we can, and any deviations are called out.

> I assume you will merge the PR since I can't do it?

@jgerigmeyer is probably better merging this and releasing; I do have permissions to merge but not to release.

jgerigmeyer:
Released in [v0.3.3](https://github.com/oddbird/popover-polyfill/releases/tag/v0.3.3)

PR #153, opened 2023-11-28
use :where for selectors

keithamus:
## Description

This wraps all selectors in a `:where()` (and also drops the redundant `:is()` which was there to lower specificity).

I tested this change on GitHub (where we use this polyfill) and I couldn't spot any regressions, styling worked as expected, and I spotted a couple of areas where we could lower our specificity. So I think this is is a worthwhile change.

## Steps to test/reproduce
This is effectively a no-op change, so no steps to reproduce.

## Show me
This is effectively a no-op change, so no steps to reproduce.

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -35,8 +35,10 @@ function patchSelectorFn<K extends string>(
 
 const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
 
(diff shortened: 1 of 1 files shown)


jgerigmeyer (approved):

mirisuzanne (approved):
great!

PR #155, opened 2023-11-29
Fix incorrect `isFocusable` result for `ShadowRoot`

### LINKED ISSUE #154: `_a.focus is not a function` with `delegatesFocus: true`, but no focusable element
So my fix for #149 works as it's supposed to. The scenario I did not account for is when there is no focusable content in the web component. In that case `focusDelegate` still returns the `ShadowRoot` and then errors in the `popoverFocusingSteps`.

I think this is due to the `TreeWalker` code where `isFocusable` falls through all the if statements and then returns `focusTarget.tabIndex !== -1` which is true for the `ShadowRoot`?

jpzwarte:
Fixes #154


keithamus (approved):
LGTM

PR #158, opened 2023-11-30
allow esc to be defaultpreventable

keithamus:
## Description

Browsers' native `popover` support allows for the `Esc` keypress to be preventable, which means that the popover won't close.

This is useful for capturing the `Esc` key behaviour from _within_ a popover to do other things without closing the popover.

Currently the polyfill does not allow for this. This PR checks the flag to ensure default hasn't been prevented.

## Steps to test/reproduce
I can follow up with steps/tests if we feel it necessary (time is short right now). 

## Show me

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -323,7 +323,11 @@ export function apply() {
   const onKeydown = (event: Event) => {
     const key = (event as KeyboardEvent).key;
     const target = event.target as Element;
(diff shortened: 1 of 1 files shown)


jgerigmeyer (approved):

keithamus:
@jgerigmeyer do you think you could release this one as a patch please?

PR #161, opened 2023-12-07
Improve ESM Documentation

wesbos:
Hello! We recently had a 500ms render blocking script load, and I think it stems back to the docs for using this being a bit confusing on how to import the polyfill. I'm clarifying how to import the polyfill with almost all JS tooling. In fact, I can't think of any tooling that doesn't support this method.


jgerigmeyer (approved):
@wesbos Thanks for the contribution! One typo, then good to merge. 🚀

Issue #176, opened 2024-02-07
Click events not correctly crossing ShadowDOM boundaries


Prestaul:
`handleInvokerActivation` is using `closest` which does not cross ShadowDOM correctly. The fix is to use `composedPath` 

```javascript
event.composedPath().find(el => el.matches && el.matches('[popovertargetaction],[popovertarget]'));
// in place of:
target.closest('[popovertargetaction],[popovertarget]')
```

In my case, we are sometimes using web-components for rendering svg icons inside of `<button>` elements. The click works if you hit the edges of the button itself (in the light), but clicking on the icons (in the shadow) does not properly trigger the popover.

PR #177, opened 2024-02-07
Ensure click events correctly cross out of ShadowDOM

Prestaul:
## Related Issue(s)
#176 

## Steps to test/reproduce
- Populate the shadow dom of any element inside of a button that is a `popovertrigger`
- When a click originates from within the shadow dom, the popover is not triggers

### DIFF
--- a/index.html
+++ b/index.html
@@ -46,6 +46,7 @@
           </ul>
         </div>
       </div>
+      <div id="popover12" popover>Popover 12</div>
       <dialog popover>I'm a dialog!</dialog>
       <div id="host"></div>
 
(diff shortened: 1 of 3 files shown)


Prestaul:
@jgerigmeyer, it looks like the build hasn't reported after almost an hour. Did I break it by force pushing? Let me know if there is anything I need to do to correct whatever I've broken.

jgerigmeyer:
> @jgerigmeyer, it looks like the build hasn't reported after almost an hour. Did I break it by force pushing? Let me know if there is anything I need to do to correct whatever I've broken.

@Prestaul Sorry about that -- as a first-time contributor I needed to approve it before actions would run.

keithamus (approved):
Nice catch. This LGTM :+1:

jgerigmeyer (approved):
Thanks!

Issue #175, opened 2024-02-07
Current implementation incompatible with css layers


Prestaul:
User-agent (and polyfilled) styles should always have lowest priority, but if a consumer uses cascade layers (i.e. `@layer`) then polyfilled popover styles have highest priority and overwrite other styles.

In any browser, the user-agent styles are effectively in the first layer and therefore any non-agent styles win, regardless of specificify. When styles include layers, styles not in a layer always win. The polyfilled styles are not in a layer and this means that, even with `:where()` and a specificity of 0-0-0-0, they always overwrite any styles on sites using cascade layers.

I'm not sure what the intended browser support matrix is for this polyfill, but all current browsers have full `@layer` support so it is possible to just drop polyfilled styles into a layer and this may be resolved. 

Another option would be to make it configurable, either with a boolean flag or possibly even a nullable layer name. The obvious drawback here is that you'd have to make a function call (like the anchor positioning polyfill) to initialize the polyfill if it has configuration.

mirisuzanne:
I chatted with @jgerigmeyer about this, and see a few options:

- A config setting (like you mentioned)
- A separate build that sets that config for you

It might make sense to provide both in this case? I imagine, for broader support, we should keep the unlayered option as the default - but make the layered option as easy an upgrade as possible.

For the details of that, I would expect we wrap the entire stylesheet with something like `@layer popover-polyfill.default { /* styles… */ }`. The double-nesting allows authors to both place `popover-polyfill` where they want it in a layer order, and also add styles to that layer above or below the defaults. Once you start layering, you may as well go all-in. :)

Curious if @keithamus has thoughts on this as well.

Prestaul:
@shannonmoeller mentioned the possibility of having a `popover-layer.min.js` as a separate build as well. I'm not sure if there is a convention here the Odd Birds prefer.

Another possibility would be an optional param added to `apply` and consumers using cascade layers would have to use the more verbose `if (!isSupported()) apply(true);` pattern of use.

keithamus:
I personally prefer zero configuration wherever possible. Given we inject the styles in JS, we could feature detect for `@layer` and do it automatically, but it may cause surprising style issues as browsers which _dont_ support `@layer` would suddenly have CSS that participates in the regular cascade in ways that perhaps we are unable to anticipate or correct for?

Having said that I don’t have enough experience in using `@layer` to know if we’d be causing gotchas with this technique. 

[...]

PR #178, opened 2024-02-07
Add support for cascade layers

Prestaul:
## Description
User-agent and polyfilled styles should always have lowest priority, but if a consumer uses cascade layers (i.e. @layer) then polyfilled popover styles have highest priority and overwrite other styles unless they are also in a layer. This change puts polyfilled styles into a layer so that consumers can control layer order.

## Related Issue(s)
#175 (original issue and discussion)

## Steps to test/reproduce
Place a style that sets the background or color of any popover element into any `@layer` block. That style will only apply correctly after this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,6 +142,11 @@ some caveats which will need accommodations:
     now is to add `!important` to conflicting properties in your `:host` rule.
     See [#147](https://github.com/oddbird/popover-polyfill/issues/147) for more.
 
+- When supported, the polyfill creates a cascade layer named `popover-polyfill`.
+  If your styles are not in layers then this should have no impact. If your
+  styles do use layers, you'll need to ensure the polyfill layer is declared
+  first. (e.g. `@layer popover-polyfill, other, layers;`)
+
 ## Contributing
 
 Visit our [contribution guidelines](https://github.com/oddbird/popover-polyfill/blob/main/CONTRIBUTING.md).
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -34,7 +34,7 @@ function patchSelectorFn<K extends string>(
 }
 
 const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
-const hasLayerSupport = Boolean(window.CSSLayerBlockRule);
+const hasLayerSupport = typeof window.CSSLayerBlockRule === 'function';
 
 // To emulate a UA stylesheet which is the lowest priority in the cascade,
 // all selectors must be wrapped in a :where() which has a specificity of zero.
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -34,10 +34,12 @@ function patchSelectorFn<K extends string>(
 }
 
 const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
+const hasLayerSupport = Boolean(window.CSSLayerBlockRule);
 
 // To emulate a UA stylesheet which is the lowest priority in the cascade,
 // all selectors must be wrapped in a :where() which has a specificity of zero.
 const styles = `
+${hasLayerSupport ? '@layer popover-polyfill {' : ''}
   :where([popover]) {
     position: fixed;
     z-index: 2147483647;
@@ -93,6 +95,7 @@ const styles = `
   :where([popover]:not(.\\:popover-open)) {
     display: none;
   }
+${hasLayerSupport ? '}' : ''}
 `;
 let popoverStyleSheet: null | false | CSSStyleSheet = null;
 export function injectStyles(root: Document | ShadowRoot) {
--- a/tests/edge-cases.spec.ts
+++ b/tests/edge-cases.spec.ts
@@ -20,3 +20,10 @@ test("removing an autoPopover from document doesn't crash page", async ({
   ).toBeUndefined();
   await expect(popover2).toBeVisible();
 });
+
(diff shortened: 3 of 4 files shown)


Prestaul:
I apologize for the stacked PR. This seemed the best way to avoid conflicts with my other PR, but I'm happy to break this up and just deal with conflicts and rebasing if one lands.

The main issue is in the `index.html` where each PR is adding a new popover and button for testing. The existing pattern of just incrementing a number in the id (i.e. `id="popover12`) doesn't lend itself to concurrent changes.

README.md
```diff
@@ -142,6 +142,11 @@ some caveats which will need accommodations:
     now is to add `!important` to conflicting properties in your `:host` rule.
     See [#147](https://github.com/oddbird/popover-polyfill/issues/147) for more.
 
+- When supported, the polyfill creates a cascade layer named `popover-polyfill`.
+  If your styles are not in layers then this should have no impact. If your
+  styles do use layers, you'll need to ensure the polyfill layer is declared
+  first. (e.g. `@layer popover-polyfill, other, layers;`)
+
```
keithamus, line 149: (Non blocking question) Can layers be re-used? Would it make sense to have `@layer polyfill {}` instead, in the hopes that it can become the defacto layer to put CSS polyfills in?

src/popover.ts
```diff
@@ -34,10 +34,12 @@ function patchSelectorFn<K extends string>(
 }
 
 const nonEscapedPopoverSelector = /(^|[^\\]):popover-open\b/g;
+const hasLayerSupport = Boolean(window.CSSLayerBlockRule);
```
keithamus, line 37: Non blocking nit:

```suggestion
const hasLayerSupport = typeof window.CSSLayerBlockRule === 'function';
```

tests/edge-cases.spec.ts
```diff
@@ -20,3 +20,10 @@ test("removing an autoPopover from document doesn't crash page", async ({
   ).toBeUndefined();
   await expect(popover2).toBeVisible();
 });
+
+test('styles in layers are winning over polyfilled styles', async ({
+  page,
+}) => {
+  const popover = (await page.locator('#popover13')).nth(0);
+  await expect(popover).toHaveCSS('background-color', 'rgb(208, 13, 30)');
+});
```
keithamus, line 29: Nice test :+1:

keithamus (approved):

README.md
Prestaul: That's actually a really good call. Yes, they can be extended and they can have hierarchy. Maybe `polyfill.popover` would be best. That gives consumers the chance to set a layer order including `polyfill` alone, but we can also set an order for specific polyfills, if needed.

@shannonmoeller also had thoughts about later naming conventions for public packages.

tests/edge-cases.spec.ts
jamesnw: Per this [blog post](https://playwrightsolutions.com/what-the-hex-or-how-i-check-colors-with-playwright), Playwright does expect it to be rgb. I'm also seeing `rgb()` in Playwright's own tests, but I'm not seeing this documented.

keithamus: I'm pretty sure this must be `getComputedStyle()`, so the browser is doing it.

Prestaul: Looks like it is spec'd that way for `getComputedStyle()`, so this is probably okay. https://stackoverflow.com/a/67006298/5628

README.md
shannonmoeller: I thought it might make sense to mirror npm or repo namespacing in the layers to ensure uniqueness and manageability. Maybe something like one of these:

- `@layer npm.oddbird.popover-polyfill`
- `@layer github.oddbird.popover-polyfill`
- `@layer vendor.oddbird.popover-polyfill`

Prestaul: Ah, I'm glad I asked, but I think that probably won't work because it would force all the `npm.*` or `oddbird.*` styles to be grouped together in the layer order. It seems reasonable to me to group all `polyfills.*` together, but even that might be presumptuous without some namespacing... 

Does anyone have thoughts about `oddbird-polyfills.popover` vs `polyfills.popover` to avoid potential naming conflicts?

shannonmoeller: Great points. I think `oddbird-polyfills.popover` is pretty unique. Though `polyfills.popover` seems good too. If people are including more than one popover polyfill on the page which would conflict with that I think they have a bigger issue to deal with.

mirisuzanne: Whatever we call our internal layers, consuming authors can create their own groupings and avoid name conflicts by nesting on their end. I don't think we should worry about trying to do those things for them. I'd say `popover-polyfill` on it's own is probably plenty, and we don't even need the sublayer I suggested in the issue.

Layer-naming-by-vendor/source conflates two things that aren't really related in my mind. But whatever we do, authors can pull these styles into an `oddbird` layer if they choose to.

mirisuzanne: Oh that's true in this case. I'm not used to injecting styles on someone else's site 😅 - but yeah, I think the simpler solution still makes sense.

jgerigmeyer (approved):
Thanks!

> The existing pattern of just incrementing a number in the id (i.e. id="popover12) doesn't lend itself to concurrent changes.

Yes -- we're working on creating a better demo page.

PR #190, opened 2024-03-22
Do not close parent popover when shadowed nested popover is opened

### LINKED ISSUE #183: Polyfill is broken when used with nested popovers and inside shadow DOM
```js
if (element.popover === "auto") {
    const originalType = element.getAttribute("popover");
    const ancestor = topMostPopoverAncestor(element) || document2;
    hideAllPopoversUntil(ancestor, false, true);
    if (originalType !== element.getAttribute("popover") || !checkPopoverValidity(element, false)) {
      return;
    }
  }
```

The above snippet is part of the `showPopover` method in the polyfill (https://github.com/oddbird/popover-polyfill/blob/main/src/popover-helpers.ts#L268). When this is called inside a nested popover inside the shadow DOM, `hideAllPopoversUntil` incorrectly hides the parent popover, which doesn't happen in browsers that already support popover.

jgerigmeyer:
See new "Toggle menu with shadowed nested popover" button with nested "Toggle inner shadowed popover" button.

Before: clicking "Toggle inner shadowed popover" closes both popovers
After: clicking "Toggle inner shadowed popover" opens inner nested popover

Fixes #183


jgerigmeyer:
@keithamus Does this seem like an okay solution to #183?

keithamus (approved):
Yeah, nice!

PR #199, opened 2024-04-24
Add support for dialog popovers.

### LINKED ISSUE #198: Polyfill doesn't seem to work with dialog element
I was trying to set a `<dialog>` to be a popover as discussed here: https://hidde.blog/popover-semantics/

However, this doesn't seem to work in Firefox; the popover does seem to work with a `<div>` element though:
https://codepen.io/codepenguy131/pen/jORdKGo

jgerigmeyer:
Fixes #198 

@keithamus @mirisuzanne I see there was already a `dialog[popover]` in the demo page, but it wasn't being used or tested, and it looks to me like this additional style change was necessary to make it work. Let me know if I'm missing some context here.

### DIFF
--- a/index.html
+++ b/index.html
@@ -72,7 +72,7 @@
         Menu with shadowed popover
         <div id="menu-host"></div>
       </div>
-      <dialog popover>I'm a dialog!</dialog>
+      <dialog id="dialog" popover>This is a dialog</dialog>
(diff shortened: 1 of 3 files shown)


keithamus (approved):
Nice!

mirisuzanne (approved):

Issue #222, opened 2024-10-04
CJS support


myasonik:
**Is your feature request related to how popovers work in general?** No

**Is your feature request related to how this polyfill works? Please describe.**
My codebase only supports CJS packages while the polyfill is only bundled for ESM. 

**Describe the solution you'd like**
Include CJS support in the package.json. esbuild seems to support it out of the box. 

**Describe alternatives you've considered**
Not sure what other options there are.

jgerigmeyer:
@myasonik Which entrypoint are you currently trying to `require` that's failing? Would #224 fix your use case?

PR #223, opened 2024-10-04
Add CommonJS build script

FelipeSilva916:
## Description
This pull request adds a CommonJS (CJS) build script to the `@oddbird/popover-polyfill` package. This change allows the polyfill to be used in environments that require CommonJS modules.
This pull request includes a small change to the `package.json` file to add a new build script for CommonJS format.

## Build script addition:
* [`package.json`](diffhunk://#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519R63): Added a new script `"build:cjs"` to bundle the source file `src/index.ts` into CommonJS format with source maps.

## Steps to test/reproduce
This pull request includes an update to the build scripts in the `package.json` file to support CommonJS module format.

Added to package.json

```
    "build:cjs": "esbuild --bundle src/index.ts --outfile=dist/popover.cjs.js --format=cjs --sourcemap",
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,6 +60,7 @@
     "build:esm": "esbuild --bundle src/index.ts --outfile=dist/popover.js --format=esm --sourcemap",
     "build:esm-min": "esbuild --bundle src/index.ts --outfile=dist/popover.min.js --format=esm --minify --sourcemap",
     "build:esm-fn": "esbuild --bundle src/index-fn.ts --outfile=dist/popover-fn.js --format=esm --sourcemap",
+    "build:cjs": "esbuild --bundle src/index.ts --outfile=dist/popover.cjs.js --format=cjs --sourcemap",
     "clean": "rm -rf dist",
     "build": "run-s clean build:* types",
     "server:start": "esbuild --bundle src/index.ts --format=esm --servedir=. --outdir=dist --sourcemap --serve=\"${PORT:-3000}\" --log-level=\"${LEVEL:-info}\"",


jgerigmeyer (commented):
@FelipeSilva916 Thanks! Because the default `index.ts` entry point has no exports, it looks like the resulting CJS build is identical to the existing ESM built, except that it has `"use strict";`. Is this actually useful for CJS environments? I was assuming #222 was requesting a CJS bundle of `index-fn.ts`, which would differ meaningfully between ESM and CJS.

cc @myasonik

I think we'll also want to document this in the README file, as well as add a `require` path to the `exports` setting in `package.json`.

package.json
```diff
@@ -19,7 +19,7 @@
     "popover"
   ],
   "type": "module",
-  "main": "./dist/popover-fn.js",
+  "main": "./dist/popover.cjs.js",
```
jgerigmeyer, line 22: This file doesn't exist. And I think CJS environments should be fine with the default `./dist/popover.js` build? I think I'd lean toward making the main build the default throughout, unless users specifically require the `/fn/` build. If we really need a CJS build here, we should create a new `popover.cjs.js` one (even though it's almost identical to `popover.js`).

```suggestion
  "main": "./dist/popover.js",
```

package.json
```diff
@@ -29,11 +29,12 @@
       "types": "./dist/index.d.ts",
       "import": "./dist/popover.js",
       "browser": "./dist/popover.iife.min.js",
-      "require": "./dist/popover.js"
+      "require": "./dist/popover.cjs.js"
```
jgerigmeyer, line 32:
```suggestion
      "require": "./dist/popover.js"
```

src/index-fn.ts
```diff
@@ -28,11 +32,13 @@ declare global {
...
 
   interface Window {
     ToggleEvent: ToggleEvent;
   }
 }
 
-export * from './popover.js';
+if (!isSupported()) {
+  apply();
+}
```
jgerigmeyer, line 44: I'm not sure I understand the changes in this file. The purpose of the `-fn` builds is for the polyfill _not_ to run automatically, but to allow users to run it conditionally on their own.

jgerigmeyer (commented):
@FelipeSilva916 Would #224 work for you?

src/index-fn.ts
FelipeSilva916: Thanks for pointing this out! You're absolutely correct—the `-fn` build should not automatically apply the polyfill.

FelipeSilva916:
> @FelipeSilva916 Would #224 work for you as an alternative?

@jgerigmeyer Let me give that a try

src/index-fn.ts
```diff
@@ -1,3 +1,7 @@
+import { apply, isPolyfilled, isSupported } from './popover.js';
+
+export { apply, isSupported, isPolyfilled };
+
```
jgerigmeyer, line 4:
```suggestion
```

src/index-fn.ts
```diff
@@ -34,5 +38,3 @@ declare global {
     ToggleEvent: ToggleEvent;
   }
 }
```
jgerigmeyer, line 40:
```suggestion
}

export * from './popover.js';
```

jgerigmeyer (commented):
@FelipeSilva916 If you revert the remaining changes to `index-fn.ts` (or explain to me why they're needed?), I'm fine merging this.

src/index-fn.ts
FelipeSilva916: @jgerigmeyer I wrote that because I believed I needed to export those in order to use it in TS, is that not the case for this?

This is my first time implementing polyfills so pardon my ignorance on this. Im learning as I go.

jgerigmeyer: No apology necessary! Thanks for explaining. In this case the `export * from './popover.js';` at the bottom of the file exports _all_ functions from `popover.js`, so this isn't needed.

FelipeSilva916: @jgerigmeyer I see what you are saying. Ok ill commit your changes on my PR, would that be all?

jgerigmeyer (approved):
Looks good -- thanks!

PR #225, opened 2024-10-08
Upgrade eslint

jgerigmeyer:


src/popover-helpers.ts
```diff
@@ -84,7 +84,6 @@ function topMostAutoPopover(document: Document): HTMLElement | null {
   const documentPopovers = autoPopoverList.get(document);
   for (const popover of documentPopovers || []) {
     if (!popover.isConnected) {
-      // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
```
jamesnw, line 87: Should we be concerned about disabling this check?

jamesnw (approved):

src/popover-helpers.ts
jgerigmeyer: I'm not sure why these got added... this rule is off by default 🤔

PR #236, opened 2025-02-26
Clean up styles for older browsers

### LINKED ISSUE #205: Demo Application Does Not Work in Safari 15 or 16
I wanted to evaluate the behavior of the polyfill in Safari 15 & 16 before integrating it into my project, so I opened the demo netlify app in each browser (I was using Saucelabs to access the browser), and in both cases, the styling of the demo app had issues:

In Safari 15, the demo application was usable, but completely unstyled.

In Safari 16, the demo application is completely unusable, as all the elements on the screen are invisible (except for the code samples, and, only when hovering, the OddBird logo in the nav bar).

It seems like, if the purpose of the library is to support older browsers, the demo app should probably work in older browsers. It certainly gives me pause as to the extent of the backwards compatible support if the demo app doesn't work in older browsers. 

How is the polyfill being tested in these browser environments? 

Note: I don't see any code for the demo app in this repo, but couldn't find a repo in your org that does have it. If there's a better place to file the ticket, please let me know.

dvdherron:
## Related Issue(s)
Fixes #205 

## Steps to test/reproduce
-  On Sauce Labs, view the demo page in Safari 15 and 16 (Ventura OS) or versions of other browsers before they implemented cascade layers. 
- Compare the the deploy preview here to the live site. In the latest versions of browsers, everything should be the same (except for the updated `@layers` popover example). 

image

### DIFF
--- a/css/demo.css
+++ b/css/demo.css
@@ -255,26 +255,6 @@ a[target='_blank']::after {
   }
 }
 
-/* for browsers that don't support popovers */
-[data-popover].\:popover-open {
-  align-items: center;
-  border: 1px solid var(--border-color);
-  box-shadow: 0 3px 3px var(--shadow);
-  color: var(--popover-color, var(--text));
-  display: flex;
-  font-family: var(--sans-serif);
-  flex-wrap: wrap;
-  gap: var(--shim);
-}
-
-@media (width <=780px) {
-  [data-popover].\:popover-open {
-    inline-size: auto;
-    justify-content: center;
-    margin-inline: var(--shim);
-  }
-}
-
 [data-popover~='layered-styles'] {
   --border-color: var(--text);
   --popover-color: var(--bg);
@@ -327,5 +307,31 @@ a[target='_blank']::after {
 #fullShadowHost {
   &:focus {
     outline: none;
+  }
+}
+
+/* for browsers that don't support popovers */
+@supports not selector(:popover-open) {
+  [popover]:not(.\:popover-open) {
+    display: none;
+  }
+
+  [data-popover].\:popover-open {
+    align-items: center;
+    border: 1px solid var(--border-color);
+    box-shadow: 0 3px 3px var(--shadow);
+    color: var(--popover-color, var(--text));
+    display: flex;
+    font-family: var(--sans-serif);
+    flex-wrap: wrap;
+    gap: var(--shim);
+  }
+
+  @media (width <=780px) {
+    [data-popover].\:popover-open {
+      inline-size: auto;
+      justify-content: center;
+      margin-inline: var(--shim);
+    }
   }
 }
--- a/css/demo.css
+++ b/css/demo.css
@@ -80,17 +80,19 @@ body {
   margin-inline: var(--page-margin);
 }
 
-@media (width >=70em) {
-  .main-header {
-    align-items: var(--header-items-align, center);
-    display: flex;
-    gap: var(--double-gap);
+@supports (container: inline-size) {
+  @media (width >=70em) {
+    .main-header {
+      align-items: var(--header-items-align, center);
+      display: flex;
+      gap: var(--double-gap);
+    }
   }
-}
 
-@media (width >=1400px) {
-  .main-header {
-    --header-items-align: last baseline;
+  @media (width >=1400px) {
+    .main-header {
+      --header-items-align: last baseline;
+    }
   }
 }
 
--- a/css/demo.css
+++ b/css/demo.css
@@ -23,10 +23,10 @@ body {
 }
 
 #banner-logo [data-logo] {
-  block-size: auto;
   height: auto;
-  inline-size: 100%;
+  block-size: auto;
   width: 100%;
+  inline-size: 100%;
 }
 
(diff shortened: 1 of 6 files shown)


dvdherron:
@stacyk I've removed the references here to layers (and most nesting). There's something happening that's causing a lot of the layout styles to not be applied. Testing on FF 95, custom properties are valid but not all of them are reflected in the styles of the page. Can you see the same thing in Safari 15?

dvdherron:
@stacyk I figured out why the size variables were not working. It's due to the `cq` units in the base sizes.

stacyk:
@dvdherron I have Safari 18 and there isn't a way to downgrade that I know of, so using Saucelabs is the way to go for that. Sounds like you are on the right trail so let me know if you want a review when ready.

dvdherron:
@jgerigmeyer @stacyk I think this is ready for review. 

I removed the references to layers (except for the demo example). I changed the example from having a red background to having a red border. With the former, an old browser could not read the text (It was white text on a white background). 

I also removed nesting and added some `@support` for container query related styles. Its not perfect in older browsers, but I'm not sure how far back we want to fully support for this page. 

If you test this page on Safari 16, it seems to only work when you choose `Ventura` as an OS version. 

The shadowed popovers in older browsers don't seem to be getting the "UA Stylesheet" styles we are injecting with JS. I think that's because we are using `.innerHTML` to create those elements @jgerigmeyer? These popovers aren't positioned so they push content around when they are opened.

dvdherron:
> Looks good to me!
> 
> > The shadowed popovers in older browsers don't seem to be getting the "UA Stylesheet" styles we are injecting with JS. I think that's because we are using `.innerHTML` to create those elements @jgerigmeyer? These popovers aren't positioned so they push content around when they are opened.
> 
> Yeah, I think this is directly related to [this note](https://github.com/oddbird/popover-polyfill/blob/main/README.md?plain=1#L140-L143). I think I fixed it (conditionally re-injecting the styles in shadow roots) in [0f301bc](https://github.com/oddbird/popover-polyfill/commit/0f301bc618ad5c7c66b511281e489360851a5fa4) -- take a look and see what you think?

Your changes fixed what I was seeing with the shadowed popovers. Thanks! This looks good to me.

css/props.css
```diff
@@ -66,3 +69,20 @@ html {
   --ccs-special--fg3: hsl(24deg 53% 27%);
   --ccs-accent--fg2: hsl(330deg 60% 35%);
 }
+
+@supports (font-size: 2cqi) {
```
stacyk, line 73: So Safari 16 supports cqi, but it looks like it doesn't like the lh unit so spacer gets ignored. Still testing Safari 15 ... 

Screenshot 2025-03-07 at 10 57 23 AM
Screenshot 2025-03-07 at 10 51 25 AM

stacyk (commented):
Super close, just a small Safari 16 issue with I think lh units (the supports query only checks for cqi)

css/props.css
dvdherron: good catch. fixed in https://github.com/oddbird/popover-polyfill/pull/236/commits/5ef1a52d3f3b1edf120ad7b6b27d287c711cdd78

Safari 16: 
https://github.com/oddbird/popover-polyfill/pull/236/commits/5ef1a52d3f3b1edf120ad7b6b27d287c711cdd78

image

image

And 15:
image

image

css/demo.css
```diff
@@ -315,3 +315,29 @@ a[target='_blank']::after {
...
+  }
+
+  [data-popover].\:popover-open {
+    align-items: center;
+    border: 1px solid var(--border-color);
+    box-shadow: 0 3px 3px var(--shadow);
+    color: var(--popover-color, var(--text));
+    display: flex;
+    font-family: var(--sans-serif);
+    flex-wrap: wrap;
```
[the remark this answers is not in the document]
dvdherron: Sorry @stacyk , missed this somehow! Fixed in https://github.com/oddbird/popover-polyfill/pull/236/commits/b7626a8ae488e1451256320f9f4fe616ad4bac5d

css/demo.css
```diff
@@ -15,27 +16,27 @@ body {
...
-  &:hover,
-  &:focus,
-  &:active {
-    transform: scale(1.05);
-  }
+#banner-logo [data-logo] {
+  block-size: auto;
+  height: auto;
+  inline-size: 100%;
+  width: 100%;
```
stacyk, line 29: @dvdherron I didn't comment on these as I think we do actually want the more "modern" property last as it will be ignored if a browser doesn't understand it. So the order:

height
block-size
width
inline-size 

is probably what we want, but other than that,  I'd say this is good to merge!

dvdherron: good catch. fixed in https://github.com/oddbird/popover-polyfill/pull/236/commits/5ebfc647404ed7cfb6db32bdf44d6ff5ce701c42 so I think this is good to merge!

dvdherron:
@stacyk Ok, I think this still needs one more thumbs up from you and then this is ready to merge.

stacyk (approved):

PR #240, opened 2025-04-07
Support for popover=hint

### LINKED ISSUE #227: popover=hint support
**Is your feature request related to how popovers work in general?**
No

**Is your feature request related to how this polyfill works? Please describe.**
Sort of, it's on adding support for a new part of the popover spec.

**Describe the solution you'd like**
Add support for `popover=hint` as described in https://github.com/whatwg/html/pull/9778.

**Describe alternatives you've considered**
Use alternative library / wait for browser support

**Additional context**
This is still early days, but having the popover polyfill support this would mean we could start using it prior to the more wider rollout in the browser engines if it's deemed stable enough.

jgerigmeyer:
Fixes #227 

This adds support for `popover=hint`, but [as pointed out](https://github.com/oddbird/popover-polyfill/issues/227#issuecomment-2687441304), it would be great to selectively polyfill `hint` in browsers that don't support it but do have basic `popover` support. #242 tracks that specifically.

Relevant links:

- https://github.com/whatwg/html/pull/9778
- https://open-ui.org/components/popover-hint.research.explainer/#popoverhint-behavior
- https://developer.chrome.com/blog/popover-hint
- https://github.com/oddbird/popover-polyfill/pull/38

### DIFF
--- a/src/popover-helpers.ts
+++ b/src/popover-helpers.ts
@@ -28,6 +28,10 @@ const combinedPopoverListForDocument = (
   return new Set([...autoPopovers, ...hintPopovers]);
 };
 
+function lastSetElement(set: Set<HTMLElement>) {
+  return [...set].pop() as HTMLElement;
+}
+
 // https://html.spec.whatwg.org/#popover-target-attribute-activation-behavior
 export function popoverTargetAttributeActivationBehavior(
   element: HTMLButtonElement | HTMLInputElement,
@@ -415,7 +419,7 @@ export function hidePopover(
   }
   const autoList = autoPopoverList.get(document) || new Set();
   const autoPopoverListContainsElement =
-    autoList.has(element) && [...autoList].pop() === element;
+    autoList.has(element) && lastSetElement(autoList) === element;
   setInvokerAriaExpanded(popoverInvoker.get(element), false);
   popoverInvoker.delete(element);
   if (fireEvents) {
@@ -425,7 +429,10 @@ export function hidePopover(
         newState: 'closed',
       }),
     );
-    if (autoPopoverListContainsElement && [...autoList].pop() !== element) {
+    if (
+      autoPopoverListContainsElement &&
+      lastSetElement(autoList) !== element
+    ) {
       hideAllPopoversUntil(element, focusPreviousElement, fireEvents);
     }
     if (!checkPopoverValidity(element, true)) {
@@ -498,13 +505,9 @@ function hidePopoverStackUntil(
     }
     if (!lastToHide) return;
     while (getPopoverVisibilityState(lastToHide) === 'showing' && set.size) {
-      hidePopover(
-        [...set].pop() as HTMLElement,
-        focusPreviousElement,
-        fireEvents,
-      );
+      hidePopover(lastSetElement(set), focusPreviousElement, fireEvents);
     }
-    if (set.has(endpoint as HTMLElement) && [...set].pop() !== endpoint) {
+    if (set.has(endpoint as HTMLElement) && lastSetElement(set) !== endpoint) {
       repeatingHide = true;
     }
     if (repeatingHide) {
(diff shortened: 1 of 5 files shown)


jgerigmeyer:
@jamesnw I pushed very little, but figured I might as well get this up since I likely won't have time to address it in the short-term and you might want to take it over while I'm out.

jgerigmeyer (commented):
Looking good! Functionally the main thing I notice missing is the special case for `hint` popovers nested inside `auto` popovers.

src/popover-helpers.ts
```diff
@@ -334,7 +401,11 @@ export function hidePopover(
     }
   }
   topLayerElements.get(document)?.delete(element);
-  autoPopoverList.get(document)?.delete(element);
+  if (element.popover === 'auto') {
+    autoPopoverList.get(document)?.delete(element);
+  } else if (element.popover === 'hint') {
+    hintPopoverList.get(document)?.delete(element);
+  }
```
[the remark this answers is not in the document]
jgerigmeyer: Adjusted in https://github.com/oddbird/popover-polyfill/pull/240/commits/9afd5aab8be95d6ffb3643521e4501f3df5217d5

src/popover-helpers.ts
```diff
@@ -135,20 +164,39 @@ function nearestInclusiveTargetPopoverForInvoker(
...
+        newPopover.popover === 'hint' ||
+        candidateAncestor.popover === 'auto'
+      ) {
+        okNesting = true;
+      }
+      if (!okNesting) {
+        candidate = candidateAncestor.parentElement;
+      }
+    }
+    candidatePosition = popoverPositions.get(candidateAncestor);
```
jgerigmeyer, line 199: I updated this to match changes to the spec.

src/popover-helpers.ts
```diff
@@ -354,10 +454,62 @@ function closeAllOpenPopovers(
...
+        foundEndpoint = true;
+      } else if (foundEndpoint) {
+        lastToHide = popover;
+        break;
+      }
+    }
+    if (!lastToHide) return;
+    while (getPopoverVisibilityState(lastToHide) === 'showing' && set.size) {
+      hidePopover(
+        [...set].pop() as HTMLElement,
```
jgerigmeyer, line 502: I'm using this technique in a couple of places to get the last item in a Set -- maybe we should abstract it out into a named function?

src/popover-helpers.ts
```diff
@@ -354,10 +454,62 @@ function closeAllOpenPopovers(
...
+    let lastToHide = null;
+    let foundEndpoint = false;
+    for (const popover of set) {
+      if (popover === endpoint) {
+        foundEndpoint = true;
+      } else if (foundEndpoint) {
+        lastToHide = popover;
+        break;
+      }
+    }
```
jgerigmeyer, line 498: I went down a rabbit hole, and this really has me stumped. The spec logic was updated in such a way that AFAICT sets `lastToHide` as the element _before_ the endpoint, whereas the prior logic set `lastToHide` as the element `after` the endpoint. The way I understand the spec, this section should be:

```suggestion
    let lastToHide = null;
    for (const popover of set) {
      if (popover === endpoint) {
        break;
      }
      lastToHide = popover;
    }
```

The problem with that is it seems to cause an infinite recursion error, because when we call `hidePopover` on the last element in the set, we end up finding the exact same `lastToHide` and calling `hidePopover` again, etc.

Is this a bug in the spec, or are we misreading it, or...? 😕

jamesnw: It looks like it is this spec bug- https://github.com/whatwg/html/issues/11007. @jgerigmeyer does that match what you were seeing?

jgerigmeyer: Yes, exactly. We are passing the set by reference so that isn't an issue for our implementation, but the loop-reverse bug is what we were hitting. Based on that issue, it seems like our current implementation is fine — at least until they change the spec. I mean we could use the spec logic and also reverse the loop, but I don’t see how that really changes anything.

jamesnw (approved):
🚀

PR #243, opened 2025-04-24
Support object options syntax

### LINKED ISSUE #241: Support `options` in `showPopover()` and `togglePopover()`
**Is your feature request related to how this polyfill works? Please describe.**

The spec has updated the arguments for `showPopover()` and `togglePopover()`. This is supported in Chrome 133 (at least partially) and Safari TP.

`showPopover()` takes an `options` object, with a `source` parameter. 
`togglePopover()` takes either a boolean `force`, or an `options` object with keys `{source, force}`

The `source` parameter has two impacts:
- improves the keyboard focus navigation order
- adds implicit anchor on the source for the popover for anchor positioning

**Describe the solution you'd like**
At the very least, `targetPopover()` should update its check, as passing an options object currently prevents a popover from showing.

I'm unclear from the spec how the navigation order is impacted. 
For the implicit anchor, we may just be able to add `anchor-name` and `position-anchor` declarations, but we'll need to think about browser support and anchor positioning polyfill impact.

**Additional context**
https://html.spec.whatwg.org/multipage/popover.html#dom-showpopover
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/showPopover#options

https://html.spec.whatwg.org/multipage/popover.html#dom-togglepopover
https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/togglePopover#parameters

jamesnw:
## Description
Adds support for passing an object to togglePopover and showPopover, but ignores the source parameter.

## Related Issue(s)
Fixes #241

## Steps to test/reproduce
I tested this locally- should we add examples of imperatively calling the functions?


src/index.ts
```diff
@@ -7,6 +7,22 @@ interface PopoverToggleTargetElementInvoker {
   popoverTargetElement: Element | null;
 }
 
+/** [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/togglePopover **/
+type PopoverTogglePopoverOptionsBoolean = boolean;
```
jamesnw, line 11: We do lose the helper that a boolean parameter is `force`, that was in the original signature `togglePopover(force: boolean)`. Is there a way to maintain that?

jgerigmeyer: Pushed in https://github.com/oddbird/popover-polyfill/pull/243/commits/33d148edb61395f1967f1655fe2d117494607f7a -- let me know if that works.

src/index.ts
jgerigmeyer, line 1: All of the changes here need to be applied to `src/index-fn.ts` as well.

jgerigmeyer (changes requested):
I want to be sure we do some functional testing on this before release -- either via an alpha release, or by setting up something so we can test from a PR deploy preview.

src/index.ts
jamesnw: Done in [904c974](https://github.com/oddbird/popover-polyfill/pull/243/commits/904c97408071cf508dd4908094b222ded8e310ef)

jamesnw:
> I want to be sure we do some functional testing on this before release -- either via an alpha release, or by setting up something so we can test from a PR deploy preview.

I set up testing from the PR deploy preview in https://github.com/oddbird/popover-polyfill/pull/243/commits/4210bcbb21a5d7b2eed74d73de92eebeffb47204- it was much easier than in Anchor Positioning.

Demo: https://codepen.io/jamessw/pen/NPPwJaY

src/index.ts
```diff
@@ -7,6 +7,22 @@ interface PopoverToggleTargetElementInvoker {
...
+  force?: boolean;
+  source?: HTMLElement;
+}
+
+export type PopoverTogglePopoverOptions =
+  | PopoverTogglePopoverOptionsBoolean
+  | PopoverTogglePopoverOptionsObject;
+
+/** [MDN Reference](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/showPopover **/
+export interface PopoverShowPopoverOptions {
```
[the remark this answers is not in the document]
jgerigmeyer: `shared-types` makes sense to me -- renamed in https://github.com/oddbird/popover-polyfill/pull/243/commits/6407b682e0a7e6456c6fe73a34fed76dcae9bc06

I also realized that the file needs to be a `.ts` file in order for `tsc` to compile it and include it with the distributed files in `dist/`.

src/popover.ts
```diff
@@ -214,13 +219,17 @@ export function apply() {
     togglePopover: {
       enumerable: true,
       configurable: true,
-      value(force: boolean) {
+      value(options: boolean | PopoverTogglePopoverOptions = {}) {
```
jgerigmeyer, line 222: While reviewing this, I realized that `togglePopover` is now supposed to return a boolean indicating whether the popover is in the "showing" state. I think we have the tools to do that -- do you want to either add that here or file a new issue?

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/togglePopover#return_value

jgerigmeyer (approved):

src/popover.ts
jamesnw: Good catch. I'm going to file a new issue, because I uncovered a minor spec deviation in the case of `Popover is not showing, and force = false`. Currently we are running `hidePopover`, but we should only be running `checkPopoverValidity`. I want to make sure there aren't significant issues there, and it may change what we'd return slightly.

PR #251, opened 2025-06-25
Respect `defaultPrevented` in `popovertarget` click handler

philipp-spiess:
Prevent `popovertarget` invokers from processing clicks when `event.defaultPrevented` is true. This makes the polyfill behave closer to the native `popovertarget` implementation where calling `event.preventDefault()` _will_ make the `popovertargetaction` not trigger.

## Test plan

Added a new e2e test to ensure this works as expected.

### DIFF
--- a/tests/triggers.spec.ts
+++ b/tests/triggers.spec.ts
@@ -259,3 +259,24 @@ test('clicking #shadowInInvoker should show then hide popover', async ({
   await page.click('#shadowInvokedHost');
   await expect(popover).toBeHidden();
 });
+
+test('clicking button[popovertarget=defaultPopover] should not trigger popover when defaultPrevented is true', async ({
+  page,
+}) => {
+  const popover = (await page.locator('#defaultPopover')).nth(0);
+  await expect(popover).toBeHidden();
+  await page.evaluate(() => {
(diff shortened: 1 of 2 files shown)


jgerigmeyer (approved):
Thanks! One minor nit; otherwise looks good to me.

tests/triggers.spec.ts
```diff
@@ -259,3 +259,24 @@ test('clicking #shadowInInvoker should show then hide popover', async ({
...
+  await page.evaluate(() => {
+    const button = document.querySelector(
+      'button[popovertarget="defaultPopover"]',
+    );
+    button?.addEventListener(
+      'click',
+      (event) => {
+        event.preventDefault();
+      },
+      { capture: true },
```
jamesnw, line 277: Is capture needed here? There shouldn't be any click event listeners on children of the button, and it looks like it works without. If it is helpful to the test, can you add a comment?

jamesnw (approved):
Thanks for submitting! Only a minor nit/question in the test.

tests/triggers.spec.ts
philipp-spiess: Yeah definitely not necessary. 😅

PR #252, opened 2025-06-25
Add position notes

jamesnw:
## Description

Adds a note that the popovers are not positioned, and recommends anchor positioning. Links to polyfill. Should it also link to something for anchor positioning? The spec, MDN?

Adds a note to the hint example that the popovers are intentionally positioned oddly.

### DIFF
--- a/index.html
+++ b/index.html
@@ -187,6 +187,17 @@
             >native Popover API support</a
           >.
         </p>
+        <p>
+          The examples on this page are not positioned. CSS anchor positioning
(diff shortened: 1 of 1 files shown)


stacyk (approved):
@jamesnw  I think what you have here is great. It addresses the general positioning, our recommendation, and a note specifically on the hint popovers.

stacyk:
>  Should it also link to something for anchor positioning? 

@jamesnw  It could on the words CSS anchor positioning (link to MDN?) but I don't think you have to.

PR #283, opened 2026-06-25
feat: add layer name option

eirikbacker:
## Prologue
First of all; thanks for a fantastic polyfill and for all the great work to make the web a little better 🙏 ☺️ 

## Description
We are building a designsystem, using layers. When using Vite dev-mode, we experience we sometimes loose control over load order of hot-reloaded CSS, and thus not always get the layer order right. Instead of fixing the order, we'd like the `popover-polyfill` to just be part of the designsystem-layer. Therefore, we introduce the `layerName` option to `apply()`


src/popover.ts
```diff
@@ -61,10 +61,10 @@ function hasLayerSupport() {
 
 // To emulate a UA stylesheet which is the lowest priority in the cascade,
 // all selectors must be wrapped in a :where() which has a specificity of zero.
-function getStyles() {
+function getStyles(layerName: string) {
   const useLayer = hasLayerSupport();
   return `
-${useLayer ? '@layer popover-polyfill {' : ''}
+${useLayer ? `@layer ${layerName} {` : ''}
```
jamesnw:
```suggestion
${useLayer ? `@layer ${CSS.escape(layerName)} {` : ''}
```
This prevents `layerName: 'foo{}'` from breaking the styles, for instance.

README.md
```diff
@@ -191,7 +191,9 @@ some caveats which will need accommodations:
 - When supported, the polyfill creates a cascade layer named `popover-polyfill`.
   If your styles are not in layers then this should have no impact. If your
   styles do use layers, you'll need to ensure the polyfill layer is declared
-  first. (e.g. `@layer popover-polyfill, other, layers;`)
+  first. (e.g. `@layer popover-polyfill, other, layers;`). You can change the
```
jamesnw, line 227: @jgerigmeyer Do you think we should we have a way to specify the layer name when it is auto-applied without calling `apply()`?

The anchor positioning polyfill supports this by setting a value on `window`- see https://github.com/oddbird/css-anchor-positioning#configuration.

src/popover.ts
```diff
@@ -152,8 +152,9 @@ export function injectStyles(root: Document | ShadowRoot) {
   }
 }
 
-export function apply() {
+export function apply(props?: { layerName?: string }) {
   if (typeof window === 'undefined') return;
+  const layerName = props?.layerName || 'popover-polyfill';
```
jamesnw:
```suggestion
  const layerName = props?.layerName ?? 'popover-polyfill';
```
I also have a slight preference for making the default name a const, with `const DEFAULT_LAYER_NAME = 'popover-polyfill'` at the top of the file.

src/popover.ts
```diff
@@ -129,8 +129,8 @@ ${useLayer ? '}' : ''}
 }
 
 let popoverStyleSheet: null | false | CSSStyleSheet = null;
-export function injectStyles(root: Document | ShadowRoot) {
-  const styles = getStyles();
+export function injectStyles(root: Document | ShadowRoot, layerName: string) {
```
jamesnw: This is also a public function- the README should document the parameter on this function as well.

jamesnw (changes requested):
Thanks for submitting this! I have a few suggestions, and I think we will want to think about how to best expose and document parameters, since this is the first one on `apply()`

Do you have a link to a Vite issue for the root cause? I did find https://github.com/vitejs/vite/issues/22705, but I'm unclear if that's the same issue.

I think the tests will pass in CI once #284  is merged. It would be nice if this had tests, but we don't have unit tests, and e2e tests won't capture this easily.

src/popover.ts
eirikbacker: oh, nice catch and new to me - thanks! ☺️

src/popover.ts
eirikbacker: Right! is there a reason the `injectStyles` function is exported? ☺️

src/popover.ts
eirikbacker: Agreed 🙌 Fixed ✅

README.md
eirikbacker: Added support for this now ☺️

eirikbacker:
Thanks for quick reply and good review @jamesnw! 🌟  I think the issue is somehow related to https://github.com/vitejs/vite/issues/4890. We experience that our consumers who load CSS through the entry HTML file in Vite have no problem, while those who load CSS in a JS-file are prone to varying loading order, even though both techniques supports hot reload. I haven't gone down the whole rabbit hole on how this happens though 😌

src/popover.ts
jamesnw: Yes- the use case from the README is this-

For browsers which don't support `new CSSStyleSheet()`, if you are building
    a ShadowRoot by setting `.innerHTML`, you'll remove the StyleSheet. Call
    `injectStyles(myShadow)` to add the styles back in

eirikbacker: arg, sorry for missing that - indeed 😌 added some info about it in the README, and moved the logic for finding layer name from argument vs. global config vs. default layer name `const` to `getStyles` to ensure same value regardless of access path ☺️

jamesnw (approved):
This looks good to me, but I'd like review from @mirisuzanne and @jgerigmeyer before merge. Thanks, @eirikbacker!

README.md
```diff
@@ -101,6 +101,38 @@ performance and robustness.
 After installation the polyfill will automatically add the correct methods and
 attributes to the HTMLElement class.
 
+## Configuration
+
+The polyfill supports a option to change the CSS layer name. When using the default version
```
jgerigmeyer:
```suggestion
The polyfill supports an option to change the CSS layer name. When using the default version
```

jgerigmeyer (approved):

mirisuzanne (approved):
looks great! Love to see layers in use, thanks for the contribution.

jgerigmeyer:
@eirikbacker Thanks! This is released in [v0.7.0](https://github.com/oddbird/popover-polyfill/releases/tag/v0.7.0).

PR #286, opened 2026-07-06
fix: allow layername dot notation

eirikbacker:
Seems like `CSS.escape` is a bit too eager as [dot notation is allowed in CSS layer names](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer#nesting_layers) ☺️

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -65,13 +65,14 @@ function hasLayerSupport() {
 // all selectors must be wrapped in a :where() which has a specificity of zero.
 function getStyles(layerName?: string) {
   const useLayer = hasLayerSupport();
-  const layerNameEscaped = CSS.escape(
+  const layerNameEscaped = (
     layerName ??
-      window.POPOVER_POLYFILL_OPTIONS?.layerName ??
-      DEFAULT_LAYER_NAME,
+    window.POPOVER_POLYFILL_OPTIONS?.layerName ??
+    DEFAULT_LAYER_NAME
   )
-    .split('\\.')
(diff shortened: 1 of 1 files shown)


src/popover.ts
```diff
@@ -69,7 +69,9 @@ function getStyles(layerName?: string) {
     layerName ??
       window.POPOVER_POLYFILL_OPTIONS?.layerName ??
       DEFAULT_LAYER_NAME,
-  );
+  )
+    .split('\\.')
+    .join('.'); // Allow dot in CSS layer name as dot notation can be used to nest layers
```
jamesnw: Thinking through the best way of doing this. The syntax is- `<layer-name> = 
  <ident> [ '.' <ident> ]*`. We want to escape the `ident`s but not the `'.'`. Would it be better to do something like this?

```js
const escaped = layerName.split('.').map(CSS.escape).join('.');
```
The edge case of a preceeding period will make the styles not apply, but I don't think we should try to handle when users pass invalid layer names.

jamesnw (approved):
Thanks for catching this and following up! I have a suggestion, but don't think it's actually impactful. 

I'm thinking there should be a native way of doing this.

src/popover.ts
eirikbacker: Smart - and cleaner! Agreed ☺️ Changed now 👍

jgerigmeyer (approved):

jgerigmeyer:
@eirikbacker Thanks!

PR #291, opened 2026-08-11
fix: improve compatibility with jsdom and happydom

eirikbacker:
- Some of our users in https://github.com/digdir/designsystemet/ are using JSDOM or HappyDOM for testing
- These testing environments are incomplete in regards of web APIs, causing issues when using popover-polyfill
- This branch fixes the issues we have encountered, including tested downstream in both JSDOM and HappyDOM
- Also fixed `HTMLElement.prototype` methods to be `writable` to align with browser native implementations

### DIFF
--- a/src/popover.ts
+++ b/src/popover.ts
@@ -22,7 +22,8 @@ export function isSupported() {
   return (
     typeof HTMLElement !== 'undefined' &&
     typeof HTMLElement.prototype === 'object' &&
-    'popover' in HTMLElement.prototype
+    'popover' in HTMLElement.prototype && // Returns true in HappyDOM
+    'showPopover' in HTMLElement.prototype // Returns false in HappyDOM
   );
 }
 
@@ -71,7 +72,7 @@ function getStyles(layerName?: string) {
     DEFAULT_LAYER_NAME
   )
     .split('.') // Allow dot as it can be used to nest layers
-    .map(CSS.escape) // But escape each part of to ensure it's safe and valid CSS
+    .map((css) => CSS.escape(css)) // But escape each part of to ensure it's safe and valid CSS (need to wrap CSS.escape in functino for JSDOM support)
     .join('.');
 
   return `
@@ -151,7 +152,8 @@ export function injectStyles(root: Document | ShadowRoot, layerName?: string) {
       popoverStyleSheet = false;
     }
   }
-  if (popoverStyleSheet === false) {
+  // JSDOM does not support adoptedStyleSheets
+  if (popoverStyleSheet === false || !root.adoptedStyleSheets) {
     const sheet = document.createElement('style');
     sheet.textContent = styles;
     if (root instanceof Document) {
@@ -215,6 +217,7 @@ export function apply(opts?: PopoverPolyfillOptions) {
     showPopover: {
       enumerable: true,
       configurable: true,
+      writable: true,
       // eslint-disable-next-line @typescript-eslint/no-unused-vars
       value(options: PopoverShowPopoverOptions = {}) {
         showPopover(this);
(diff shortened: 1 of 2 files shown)


src/popover.ts
```diff
@@ -71,7 +72,7 @@ function getStyles(layerName?: string) {
     DEFAULT_LAYER_NAME
   )
     .split('.') // Allow dot as it can be used to nest layers
-    .map(CSS.escape) // But escape each part of to ensure it's safe and valid CSS
+    .map((css) => CSS.escape(css)) // But escape each part of to ensure it's safe and valid CSS (need to wrap CSS.escape in functino for JSDOM support)
```
jamesnw:
```suggestion
    .map((css) => CSS.escape(css)) // But escape each part of to ensure it's safe and valid CSS (need to wrap CSS.escape in function for JSDOM support)
```

src/popover.ts
```diff
@@ -215,6 +217,7 @@ export function apply(opts?: PopoverPolyfillOptions) {
     showPopover: {
       enumerable: true,
       configurable: true,
+      writable: true,
```
jamesnw, line 220: I confirmed this matches browser implementation/spec

src/popover.ts
```diff
@@ -395,7 +400,8 @@ export function apply(opts?: PopoverPolyfillOptions) {
       target &&
       (key === 'Escape' || key === 'Esc')
     ) {
-      hideAllPopoversUntil(target.ownerDocument, true, true);
+      const owner = target.ownerDocument || target;
+      hideAllPopoversUntil(owner, true, true);
```
jamesnw, line 405: Can you add a comment about this change?

jamesnw (changes requested):
Thanks for submitting this! A few minor questions/changes. I'm also unclear on the intention for HappyDOM/JSDOM- is it that the polyfill is applied, or that it isn't applied? Either way, I think that would be a good note to add to the README.

src/popover.ts
eirikbacker: Done :)

eirikbacker:
> applied

We are distributing a designsystem, where your polyfill is automatically applied if needed :) Some of our consumers are making quite extensive tests around their applications in JSDOM/HappyDOM, and naturally want to simulate opening dropdowns etc. Therefore, making the polyfill run smoothly also in these fake DOM environments helps us provide a smooth experience both for end users and when testing :)

src/popover.ts
```diff
@@ -395,7 +400,9 @@ export function apply(opts?: PopoverPolyfillOptions) {
       target &&
       (key === 'Escape' || key === 'Esc')
     ) {
-      hideAllPopoversUntil(target.ownerDocument, true, true);
+      // target can evaluates to "document" in JSDOM, causing owerDocument to be undefined
```
jgerigmeyer:
```suggestion
      // target evaluates to "document" in JSDOM, causing ownerDocument to be undefined
```

src/popover.ts
```diff
@@ -395,7 +400,9 @@ export function apply(opts?: PopoverPolyfillOptions) {
...
+      const owner = target.ownerDocument || target;
```
jgerigmeyer, line 404: Let's apply this same defensive strategy to the `target.ownerDocument` call in `lightDismissOpenPopovers` (`src/popover-helpers.ts:568`)

jgerigmeyer (approved):
A few nits, but otherwise this looks fine to me!

src/popover.ts
eirikbacker: Good point, done :)

jamesnw (approved):
LGTM- Thanks!

PR #293, opened 2026-09-21
fix: SSR compatibility

eirikbacker:
We have gotten some feedback from users in non-Node.js environments (i.e. Java emulating JavaScript, used for React server side rendering) saying their apps crash with a "Event is undefined" message. Patching this with a fallback here, following the pattern from other files in the polyfill ☺️


src/events.ts
```diff
@@ -3,7 +3,8 @@ export interface ToggleInit extends EventInit {
   newState: string;
 }
 
-export class ToggleEvent extends Event {
+const EventClass = globalThis.Event || class {};
+export class ToggleEvent extends EventClass {
```
jgerigmeyer: Nit: I think `BaseEvent` is a slightly clearer name here.

```suggestion
const BaseEvent = globalThis.Event || class {};
export class ToggleEvent extends BaseEvent {
```

jgerigmeyer (approved):
Thanks!

